PR #7, opened 2021-03-09
Add partition for chrysalis

tangq:
Chrysalis recently created the debug queue. Add partition to make use of it.


golaz:
@tangq : does this still work if a user doesn't specify a partition in their cfg file, and partition is set to "" from default.ini?

tangq:
@golaz , I got errors when not specifying the partition and using the default "".

How does partition work on other machines? Does the default work?

golaz:
I don't remember, but it probably doesn't work either. We could change the header to not add the line with --partition if it is not defined. Not sure if it's worth it.

PR #8, opened 2021-03-16
Updating mpas_analysis for maint-1.0 output

xuezhengllnl:
The changes enable mpas_analysis on maint-1.0 simulations with the corrected computation of MOC. 

The setting for maint-1.0 in cfg:
mpaso_nml = "mpas-o_in"
mpassi_nml = "mpas-cice_in"
stream_ocn = "streams.ocean"
stream_ice = "streams.cice"
PostMOC = True


templates/mpas_analysis.bash
```diff
@@ -271,6 +271,17 @@ endYear = ${tsY2}
 startYear = ${ensoY1}
 endYear = ${ensoY2}
 
+[streamfunctionMOC]!
```
golaz, line 274: Any reason for the exclamation mark ! at the end of lines 274-282?

golaz (approved):
Looks good to me. This will be useful.

templates/mpas_analysis.bash
xuezhengllnl: I removed the exclamation marks.

PR #18, opened 2021-03-29
Add ts options for high-frequency model output

tangq:
The ncclimo command requires day-per-file (dpf) and time-per-day (tpd) information to correctly handle high-frequency model output.

Tested successfully on compy with the config file: /qfs/people/tang338/E3SM/utils/zppy/20210326_chemUCI.beta1_01.amip.ne30pg2_r05_oECv3.compy.cfg


tangq:
Details about the ncclimo discussion can be found here: https://github.com/nco/nco/issues/237

golaz (changes requested):
@tangq : thanks for tracking this down. The changes look reasonable. There is only one small change to make.

Under `[[_many__]]`, the variables should be added like you did, but the default values should always be 'None'. This is needed for inheritance to work properly.

tangq:
@golaz , I made the suggested changes. Merge to main now.

PR #19, opened 2021-03-29
skip overlapped data from multiple moc files in global_time_series/coupled_global.py

xuezhengllnl:
function getmoc() in global_time_series/coupled_global.py has been updated to skip the overlapping among multiple mocTimeseries files

### DIFF
--- a/global_time_series/coupled_global.py
+++ b/global_time_series/coupled_global.py
@@ -24,7 +24,7 @@ def getmoc(dir_in):
     var0 = fin['mocAtlantic26'][:]
     for iyear in range(np.int(time0[0]),np.int(time0[-1])+1):
         if(i>0 and iyear <= time[-1]):
-          #print('the amoc value for year',iyear, 'has been included in the moc time series from another moc file',files[i-1], time[-1], 'Skipping...')
+          print('the amoc value for year',iyear, 'has been included in the moc time series from another moc file',files[i-1], time[-1], 'Skipping...')
         else:
           imon = np.where(time0==iyear)[0]
           if(len(imon)==12):
--- a/global_time_series/coupled_global.py
+++ b/global_time_series/coupled_global.py
@@ -23,12 +23,16 @@ def getmoc(dir_in):
     time0=fin['year'][:]
     var0 = fin['mocAtlantic26'][:]
     for iyear in range(np.int(time0[0]),np.int(time0[-1])+1):
-     imon = np.where(time0==iyear)[0]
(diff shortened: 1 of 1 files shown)


global_time_series/coupled_global.py
```diff
@@ -23,12 +23,16 @@ def getmoc(dir_in):
...
     var0 = fin['mocAtlantic26'][:]
     for iyear in range(np.int(time0[0]),np.int(time0[-1])+1):
-     imon = np.where(time0==iyear)[0]
-     if(len(imon)==12):
-      var=np.append(var,np.mean(var0[imon]))
-      time=np.append(time,iyear)
-     else:
-      print('error in input file :',files[i])
+        if(i>0 and iyear <= time[-1]):
+          #print('the amoc value for year',iyear, 'has been included in the moc time series from another moc file',files[i-1], time[-1], 'Skipping...')
```
forsyth2, line 27: You need to uncomment this or else Python will fail with `IndentationError: expected an indented block`.

Also, it would be better to format as:
`print("The amoc value for year {} has been included in the moc time series from another moc file {} {}. Skipping...".format(iyear, files[i-1], time[-1]))`

global_time_series/coupled_global.py
```diff
@@ -102,7 +106,7 @@ def add_trend(year, var, year1, year2, ax, format="%4.2f", lw=1, color='b',
    'vol':'{}/post/ocn/glb/ts/monthly/10yr/glb.xml'.format(case_dir),
    'name': experiment_name,
    'yoffset':0.0,
-   'yr':([21,100],),
+   'yr':([1,100],),
```
forsyth2, line 109: @golaz Do we want the default `yr` to start at 1 or keep it at 20? We currently expect the user to change the `exps` variable according to what experiment they want to run. Perhaps it would be better to read in a text file of experiments the user wants to run?

forsyth2 (changes requested):
@xuezhengllnl Thanks for the edit! I'm unfamiliar with the problem you're fixing here though. Have you tested this change and confirmed it works?

global_time_series/coupled_global.py
golaz: Changing to year 1 is fine. For now, the script will require manual editing.

golaz (approved):

golaz:
@xuezhengllnl, sorry I didn't realize I had not approved this yet. @forsyth2, I think we can merge this. Eventually we will have a new task to fully automate this/

PR #2, opened 2021-02-24
Update to add more diagnostic sets

forsyth2:
Update to add more diagnostic sets.

PR #23, opened 2021-04-02
Print e3sm diags output

### LINKED ISSUE #13: Output of `e3sm.py` is not printed
Currently, if `python e3sm.py` fails, the error may not be printed to the `e3sm_diags_*.o*` log file. However, if `python e3sm.py` is run from the terminal, the full error stack trace will print. We therefore have to get all otuput of `python e3sm.py` to print in the log file

forsyth2:
Remove buffering to ensure errors are printed in the log file. Resolves #13.

### DIFF
--- a/templates/e3sm_diags.bash
+++ b/templates/e3sm_diags.bash
@@ -187,9 +187,9 @@ EOF
 cat > e3sm_diags.cfg << EOF
 {% include cfg %}
 EOF
-command="python e3sm.py -d e3sm_diags.cfg"
+command="python -u e3sm.py -d e3sm_diags.cfg"
 {% else %}
-command="python e3sm.py"
+command="python -u e3sm.py"
 {% endif %}
 
 # Run diagnostics


forsyth2:
@golaz This is ready for review. Note that you only have to review the last commit (https://github.com/E3SM-Project/zppy/pull/23/commits/f57a7ace605153afbbb6ab7e899773663746ac2c). The first four commits are from #2.

The actual change is to add `-u` which turns off buffering. I was concerned this would slow down the task, but this doesn't appear to be the case.

The alternative option is what I've commented out in that commit: using `flush()`. Unfortunately, that option seems to only work occasionally, whereas `-u` will consistently print the full error stack trace.

templates/default.ini
```diff
@@ -66,7 +66,7 @@ qos = string(default="regular")
 nodes = integer(default=1)
 walltime = string(default="02:00:00")
 mapping_file = string(default="")
-vars = string(default="FSNTOA,FLUT,FSNT,FLNT,FSNS,FLNS,SHFLX,QFLX,TAUX,TAUY,PRECC,PRECL,PRECSC,PRECSL,TS,TREFHT,CLDTOT,CLDHGH,CLDMED,CLDLOW")
+vars = string(default="FSNTOA,FLUT,FSNT,FLNT,FSNS,FLNS,SHFLX,QFLX,TAUX,TAUY,PRECC,PRECL,PRECSC,PRECSL,TS,TREFHT,CLDTOT,CLDHGH,CLDMED,CLDLOW,U")
```
golaz, line 69: Minor detail I should have caught earlier: we don't need globally averaged U by default, you can remove it here. But we do need it above for the regridded time series.

templates/e3sm_diags.bash
```diff
@@ -94,9 +190,9 @@ EOF
 cat > e3sm_diags.cfg << EOF
 {% include cfg %}
 EOF
-command="python e3sm.py -d e3sm_diags.cfg"
+command="python -u e3sm.py -d e3sm_diags.cfg"
 {% else %}
-command="python e3sm.py"
+command="python -u e3sm.py"
```
golaz, line 195: Nice! I didn't know about this command line option for unbuffered output.

golaz (approved):
Looks good.

PR #25, opened 2021-04-21
Create webpage for simulation

### LINKED ISSUE #24: Automatically create information page for simulation
Make it so `zppy` will create an information web page for a simulation, so that users don't have to create a Confluence page manually. This issue could be resolved in steps, until the generated page looks good. It would be good for the page to use Bokeh to have interactive time series plots.

forsyth2:
Create a webpage for a simulation automatically as part of `post.py`. Resolves #24.


post.py
```diff
@@ -79,3 +80,32 @@
 # mpas_analysis tasks
 mpas_analysis(config, scriptDir)
 
+# Create page for the simulation
```
forsyth2, line 83: Might be useful to add variables for `generate_global_time_series_plots.sh` to the configuration file, so it can be run automatically from `post.py`. Then we can also find the global time series link to display.

post.py
```diff
@@ -79,3 +80,32 @@
...
+    num_years = end_yr - start_yr + 1
+    viewer = "{}/viewer".format(os.path.join(time_periods_dir,time_period))
+    if num_years in viewer_dir.keys():
+      viewer_dir[num_years].append(viewer)
+    else:
+      viewer_dir[num_years] = [viewer]
+  for num_years in viewer_dir.keys():
+    viewers = sorted(viewer_dir[num_years])
+    for viewer in viewers:
+      print(viewer)
```
forsyth2, line 103: Still need to translate file paths to web links.

post.py
```diff
@@ -79,3 +80,33 @@
...
+if config['mpas_analysis']['active']:
+  mpas_dir = os.path.join(analysis_dir, 'mpas_analysis')
+  pages = os.listdir(mpas_dir)
+  pages = ["{}".format(os.path.join(analysis_html, 'mpas_analysis', page)) for page in pages]
+  for page in pages:
+    print(page)
+
+# TODO: create a nice webpage (using Pandoc?) rather than just printing links
```
forsyth2, line 112: Could use pandoc (https://pandoc.org/index.html)

post.py
forsyth2: See #26 for the global time series plot generation

sarats:
@chengzhuzhang I would like to discuss sometime if PACE integration can help in this regard.

chengzhuzhang:
@sarats Thank you for following this! I understand this issue particularly targets automatically creating webpages including analysis results. Integration with PACE will enable inclusion of performance and run es3m info. That would be really nice, and could be a final goal. Happy to talk more on this.

golaz:
My goal for zppy has always been to free model runners from all the tedious tasks that come along with running, analyzing, documenting simulations. I think we have made great progress, but we still a number of tedious tasks.

One of them is creating a Confluence page documenting the simulation: listing all the simulation details, jobs and creating links to all the analysis figures. So I like the idea of a more coherent web design. In addition to being more aesthetically pleasing, it could also alleviate some manual labor.

@wlin7 has also developed a very nice integrated viewer for E3SM Diags that allows for easy comparison of simulations. I'd like to see it integrated somehow.

Lots of ideas we could consider, but we should proceed carefully and thoughtfully. I don't want zppy to become overly complicated or rigid.

PR #45, opened 2021-05-18
Add tests

### LINKED ISSUE #16: Add testing
Add a unit test that runs on Chrysalis or another machine (not on GitHub Actions Linux machine).

forsyth2:
Add tests. Resolves #16.

- [x] Test bash file diffs.
- [x] Test image diffs for E3SM Diags, MPAS Analysis, and global time series plots.
- [x] Make sure `test_subsections.py` works.
- [ ] Test using `environment_commands` (e.g., `qbo` for 50 years, which requires the latest E3SM Diags).
- [x] Be able to run all tests with `python -m unittest`.


tests/test_complete_run.cfg
```diff
@@ -0,0 +1,98 @@
+[default]
+input = /lcrc/group/e3sm/ac.forsyth2/E3SM_simulations/20210122.v2_test01.piControl.ne30pg2_EC30to60E2r2-1900_ICG.chrysalis
```
forsyth2, line 2: I was using my own Chrysalis run as input. Is there another input that would be a better test case?

tests/test_complete_run.cfg
```diff
@@ -0,0 +1,98 @@
...
+input_subdir = archive/atm/hist
+output = /lcrc/group/e3sm/ac.forsyth2/zppy_test_output/20210122.v2_test01.piControl.ne30pg2_EC30to60E2r2-1900_ICG.chrysalis
```
forsyth2, line 4: These files are too big to store in the `zppy` repo. Should we just tell users to edit this file to set their own `output` and `www`? (That will require some modifications to `test_complete_run.py`).

tests/test_complete_run.py
```diff
@@ -0,0 +1,105 @@
...
+            draw.rectangle(((left, upper), (right, lower)), outline="red")
+            diff.save(
+                os.path.join(diff_dir, "{}_diff.png".format(simple_image_name)),
+                "PNG",
+            )
+
+
+class TestCompleteRun(unittest.TestCase):
+    def test_complete_run(self):
+        # Run `zppy -c test_complete_run.cfg` prior to running this test!!!
```
forsyth2, line 74: I don't see a good way for the Python test to actually run the `zppy -c` command. (We could have the test run the command and then just keep checking if the status files are all finished before moving on to the actual test.)

tests/test_complete_run.py
```diff
@@ -0,0 +1,105 @@
...
+        expected_images_dir = "/lcrc/group/e3sm/public_html/zppy_test_resources/expected_complete_run"
+
+        mismatched_images = []
+
+        counter = 0
+        with open(expected_images_file) as f:
+            for line in f:
+                counter += 1
+                if counter % 250 == 0:
+                    print('On line #', counter)
```
forsyth2, line 88: There are 3,879 images that are checked as part of this test.

zppy/amwg.py
```diff
@@ -49,10 +49,11 @@ def amwg(config, scriptDir):
             # List of dependencies
             dependencies = [ os.path.join(scriptDir, 'climo_%s_%04d-%04d.status' % (c['grid'],c['year1'],c['year2'])), ]
 
-            # Submit job
-            jobid = submitScript(scriptFile, dependFiles=dependencies, export='NONE')
+            if not c['dry_run']:
```
forsyth2, line 52: This file is never tested with `active = True`.

zppy/e3sm_diags_vs_model.py
```diff
@@ -52,10 +52,11 @@ def e3sm_diags_vs_model(config, scriptDir):
             # List of dependencies
             dependencies = [ os.path.join(scriptDir, 'climo_%s_%04d-%04d.status' % (sub,c['year1'],c['year2'])), ]
 
-            # Submit job
-            jobid = submitScript(scriptFile, dependFiles=dependencies, export='NONE')
+            if not c['dry_run']:
```
forsyth2, line 55: This file is never tested with `active = True`.

zppy/templates/default.ini
```diff
@@ -79,7 +80,7 @@ diff_title = string(default="Model - Observations")
 run_type = string(default="model_vs_obs")
 tag = string(default="model_vs_obs")
 output_format = string_list(default=list("png"))
-output_format_subplot = string_list(default=list(""))
+output_format_subplot = string_list(default=list())
```
forsyth2, line 83: This turns off subplots by default

forsyth2 (commented):
@golaz This is (mostly) ready for review.

The only thing left to do is to test `environment_commands`. Since the slowdown associated with `e3sm_diags_env` is proving difficult to debug, I might just do that as another pull request.

To run all tests, run `python -m unittest tests/*.py`. This takes about 7 minutes to run all tests. (Run `conda env create -f conda/dev.yml` to create the `zppy_dev` environment).

As with the complete run test for E3SM Diags, I placed large directories used for testing in their own space: `/lcrc/group/e3sm/public_html/zppy_test_resources`.

Since some changes are in the `zppy` directory, we should merge this before we release v1.0.0.

tests/test_complete_run.cfg
golaz: This is fine for now. We could update later when we have official v2 simulations.

tests/test_complete_run.py
golaz: Let's do like this. Checking for completion would be cumbersome indeed.

zppy/amwg.py
golaz: Yes, this is a user-contributed diagnostics.

zppy/e3sm_diags_vs_model.py
golaz: Maybe this could be done later.

zppy/templates/default.ini
golaz: Thanks for fixing this.

tests/test_complete_run.cfg
golaz: That's probably the only option.

tests/test_complete_run.py
golaz: Wow.

golaz (approved):
I have not tried running this myself (let me know if you'd like me to). You can go ahead and merge.

PR #48, opened 2021-05-22
Fixed inconsistencies

golaz:
Fixed some inconsistencies in default.ini that were causing problems wit the validator.


forsyth2 (approved):
I haven't run with these changes, but these look good to me. Thanks!

PR #63, opened 2021-05-31
Add MPAS-Analysis parameters for mapping in parallel

xylar:
The new E3SM-Unified supports ESMF map-file generation in parallel with system MPI on most machines.  (ESMF could not be built with system MPI on Cori.)  On Anvil and Chrysalis, MPAS-Analysis will not be able to run unless ESMF calls are made with `srun` (which is not the MPAS-Analysis default).

This merge updates the config options related to parallel map-file generation to be consistent with the new MPAS-Analysis and E3SM-Unified.

PR #64, opened 2021-05-31
Include MPAS-Analysis ncclimo options in zppy config

xylar:
Users may wish (or need) to change these config options in order for `ncclimo` to work for large MPAS-Ocean and -Seaice meshes.

Related to #9


golaz (approved):

xylar:
@forsyth2, please don't merge this PR until after #63.  Otherwise, you will also merge #63, maybe without meaning to.

xylar:
@forsyth2, these commits cannot be squashed.  Rather, the first commit is he only commit from #63.  This PR branched off from #63 because it would otherwise conflict with it.  So to have this PR have a clean commit history, you need to merge #63, then I need o rebase this onto the new `main`, then you can merge this (which will now have a single commit in the PR).  Alternatively, we can live with the fact that the commit from #63 is also listed in this PR.  That will speed up the process but leave GitHub a little less tidy.

forsyth2:
@xylar I just merged #63, so you're good to squash these commits and merge. Thanks!

xylar:
@forsyth2, okay, rebased!

forsyth2 (approved):

PR #65, opened 2021-05-31
Update the default parallelTaskCount in MPAS-Analysis

xylar:
For all but the largest meshes, 12 is better default than 6.


golaz (approved):

forsyth2 (approved):

xylar:
This one has also been rebased to fix the conflict.

forsyth2:
> This one has also been rebased to fix the conflict.

@xylar I just merged #64 and now this is showing a conflict again. I don't see why there would be a conflict though. I guess just rebase again?

xylar:
...and rebased again.  That's what I get for editing the same code in 3 or 4 places...

PR #68, opened 2021-06-01
Simplify the MPAS-Analysis config file

xylar:
A lot of config options are included in the generated config file but are not different from the defaults, nor are the configurable through the zppy config file, nor are they config options that a user would be likely to change by hand (which doesn't seem to be the way zppy is supposed to work anyway).


golaz (approved):
The template mpas_analysis.bash is based on an older cfg file that I was using. Very glad to see you are willing to go through it and clean it up as I'm not the right person for doing so.

xylar:
More than happy to help, now that I understand zppy better.

PR #75, opened 2021-06-16
fix test name for diurnal cycle e3sm-diags run

chengzhuzhang:
This PR fixes an issue that diurnal cycle in e3sm-diags runs won't run. i.e. ""OSError: No file found for 20210603.v2rc3c.piControl.naRRM and JJA in climo_atm_monthly_diurnal_8xdaily_180x360_aave"" where `20210603.v2rc3c.piControl.naRRM` is a user specified `short_name`, but `test_name` should be used to search for required diurnal cycle climatology files.


chengzhuzhang:
@forsyth2 this should fix the diurnal cycle in RRM run. In the piControl run, short name = case name, so there was not an issue.

forsyth2 (approved):
@chengzhuzhang Thanks, this should be good to merge. I'll test it along with some other fixes I'll be merging soon.

PR #83, opened 2021-06-23
Automate docs

forsyth2:
Automate docs. Resolves #51.

This is technically for `v1.0.0`. However, since none of the changes should affect `zppy` users, it should be fine to merge after the release of `v1.0.0`.

PR #76, opened 2021-06-16
Docs for testing E3SM Unified

forsyth2:
Docs for testing E3SM Unified

PR #84, opened 2021-06-24
Docs for testing E3SM Unified

forsyth2:
Docs for testing E3SM Unified. Replaces #76. (Now that #83 merged, docs should be merged to `main` rather than `gh-pages`).

### DIFF
--- a/docs/source/dev_guide/testing_e3sm_unified.rst
+++ b/docs/source/dev_guide/testing_e3sm_unified.rst
@@ -1,0 +1,366 @@
+.. _testing_e3sm_unified:
+
+*************************************
+Testing zppy for E3SM Unified release
+*************************************
+
+This tutorial covers testing ``zppy`` for an E3SM Unified release.
+Examples are from testing for E3SM Unified v1.5.0.
+
+Parameters
+==========
+
+Only a few parameters in the configuration file need to change for different machines.
+Be sure to change the file paths to fit your own workspaces.
+
+.. warning::
+
+    Also consider the space quotas or purge policies for the file paths you use.
+    You'll be storing about 2 TB of data.
+
+Chrysalis:
+
+- <INPUT>: /lcrc/group/e3sm/ac.forsyth2/E3SM_simulations/e3sm_unified_test_simulation/20210528.v2rc3e.piControl.ne30pg2_EC30to60E2r2.chrysalis
+- <OUTPUT>: /lcrc/group/e3sm/ac.forsyth2/E3SM_simulations/e3sm_unified_test_simulation/
+- <WWW>: /lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2/E3SM/e3sm_unified_test_simulation/
+- <PARTITION>: compute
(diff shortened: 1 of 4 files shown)


forsyth2:
@chengzhuzhang @golaz This is ready for review. You can see what the docs look like at https://portal.nersc.gov/project/e3sm/forsyth/pr84_build/html/dev_guide/testing_e3sm_unified.html. Let me know if any steps need clarification and/or corrections. Thanks!

docs/source/dev_guide/testing_e3sm_unified.rst
```diff
@@ -0,0 +1,361 @@
...
+- <PARTITION>: slurm
+- <ENVIRONMENT_COMMANDS>: source /share/apps/E3SM/conda_envs/test_e3sm_unified_1.5.0rc8_compy.sh
+- <MAPPING_FILE>: /compyfs/zender/maps/map_ne30pg2_to_cmip6_180x360_aave.20200201.nc
+- <REFERENCE_DATA_PATH>: /compyfs/e3sm_diags_data/obs_for_e3sm_diags/climatology
+- <OBS_TS>: /compyfs/e3sm_diags_data/obs_for_e3sm_diags/time-series
+- <DC_OBS_CLIMO>: /compyfs/e3sm_diags_data/obs_for_e3sm_diags/climatology
+
+Cori:
+
+- <INPUT>: /global/cscratch1/sd/forsyth/e3sm_unified_test_zppy/20210528.v2rc3e.piControl.ne30pg2_EC30to60E2r2.chrysalis
```
chengzhuzhang, line 42: The input data on Cori are  saved on scratch system, which has a purge policy (12 weeks). To be safe, maybe we should move them to somewhere more permanently, i.e. under cfs file system?

chengzhuzhang (commented):
This is very helpful. I tested on Cori and works well. I only left a comment on possibly moving input data to a more permanent directory on Cori.

docs/source/dev_guide/testing_e3sm_unified.rst
forsyth2: Do we anticipate keeping this data around? It seems like it's only useful for testing -- i.e., we could just delete it after E3SM Unified is released.

Furthermore, doesn't `cfs` have a significantly smaller space quota than `cscratch`?

chengzhuzhang: I think this dataset is useful for testing diagnostics tools on v2 data. It's helpful to keep it. What's the size of the directory?

forsyth2: Ah ok that makes sense. `/global/cscratch1/sd/forsyth/e3sm_unified_test_zppy/20210528.v2rc3e.piControl.ne30pg2_EC30to60E2r2.chrysalis` is 2 terabytes in size.

chengzhuzhang: hmm, it's bigger than i thought. I found that cfs  and scratch have the same space quota (20T).  but I guess either way is fine. We can always transfer if the data are being purged.

forsyth2: I can add a note in the documentation for testers to think about what file path is best to store their data.

PR #97, opened 2021-07-14
Shorten arguments for ncclimo

### LINKED ISSUE #96: Command line arguments too long
Some climo operations with zppy fail (for example 200 year diurnal climo) because the command line arguments passed to ncclimo exceed the maximum length set by the OS.

@czender recommends an alternative way of passing input files:

```
srun -N 1 --time=30 --pty bash
caseid='20210625.v2rc3c-GWD.piControl.ne30pg2_EC30to60E2r2.chrysalis.eam.h4'
drc_in='/lcrc/group/e3sm/ac.golaz/E3SM_simulations/20210625.v2rc3c-GWD.piControl.ne30pg2_EC30to60E2r2.chrysalis/archive/atm/hist'
drc_out=${DATA}/ne30/clm
drc_rgr=${DATA}/ne30/rgr
yr_srt=801
yr_end=1000
cd ${drc_in};ls ${caseid}.0[89]??-*.nc ${caseid}.1000-*-*.nc > ~/input.txt
cd ${drc_in};ncclimo --dbg=1 --parallel=bck --climatology_mode=hfc --case=${caseid} --vars=PRECT --yr_srt=${yr_srt} --yr_end=${yr_end} --map=${DATA}/maps/map_ne30pg2_to_cmip6_180x360_aave.20200201.nc --output=${drc_out} --regrid=${drc_rgr} < ~/input.txt > ~/ncclimo.out 2>&1 & 
```

The key is to create a file containing a list of input files (`input.txt`) then feed that list as stdin (`< ~/input.txt`) to `ncclimo`. Very clever.

We need to modify climo.bash (diurnal section), as well as ts.bash which could have the same issue.

forsyth2:
Shorten arguments for `ncclimo` calls related to diurnal cycle. Resolves #96.

### DIFF
--- a/zppy/templates/climo.bash
+++ b/zppy/templates/climo.bash
@@ -83,8 +83,9 @@ then
 fi
 {%- endraw %}
 
+ls {{ case }}.{{ input_files }}.????-*.nc > input.txt
 # Now, call ncclimo
-ncclimo \
+cat input.txt | ncclimo \
 --case={{ case }}.{{ input_files }} \
 {%- if exclude %}
 -n '-x' \
@@ -103,8 +104,7 @@ ncclimo \
 --output=trash \
 --regrid=output \
 {%- endif %}
---climatology_mode=hfc \
-{{ case }}.{{ input_files }}.????-*.nc
+--climatology_mode=hfc
 
 {% else %}
 # --- Unsupported climatology mode ---
--- a/zppy/templates/ts.bash
+++ b/zppy/templates/ts.bash
@@ -56,8 +56,9 @@ fi
 {%- endraw %}
 {%- endif %}
 
+ls {{ case }}.{{ input_files }}.????-*.nc > input.txt
 # Generate time series files
-ncclimo \
+cat input.txt | ncclimo \
 -c {{ case }} \
 -v {{ vars }} \
 --yr_srt={{ yr_start }} \
@@ -79,7 +80,6 @@ ncclimo \
 --dpf={{ dpf }} \
 --tpd={{ tpd }} \
 {%- endif %}
-{{ case }}.{{ input_files }}.????-*.nc
 
 if [ $? != 0 ]; then
   cd {{ scriptDir }}


forsyth2:
@golaz I'm not sure I did the fix correctly. I'm getting the following errors:

In `climo_atm_monthly_diurnal` output:
```
ncks: ERROR file time,1 not found. It does not exist on the local filesystem, nor does it match remote filename patterns (e.g.\
, http://foo or foo.bar.edu:file).
ncks: HINT file-not-found errors usually arise from filename typos, incorrect paths, missing files, or capricious gods. Please\
 verify spelling and location of requested file. If the file resides on a High Performance Storage System (HPSS) accessible vi\
a the 'hsi' command, then add the --hpss option and re-try command.
ncclimo: ERROR Inferred value of tpd= is not a positive integer. User requested high-frequency climatology or splitter operati\
ons without specifying the number of timesteps per day (tpd) in the input files, so ncclimo tried to infer tpd from the first \
input file. This operation resulted in a nonsense value. This is not surprising since a number of assumptions must hold true f\
or the inferral to work correctly.
HINT: Explicitly specify the number of timesteps-per-day in every input file with the --tpd option.
```

In all `ts_*` output:
```
ncclimo: ERROR Specify input file(s) with -i $drc_in or with positional argument(s) or with stdin
ncclimo: HINT Pipe input file list to stdin with, e.g., 'ls *.nc | ncclimo'
```

golaz:
Looks like it's not working. Maybe rerun with the debug option and check the content of input.txt.

forsyth2:
@golaz I added a second commit that seems to work when I run `zppy` with it. Let me know if this change looks good and I'll merge.

golaz:
@forsyth2 : interesting how the change from `< input.txt` to `cat input.txt |` solves your problem. Some OS subtleties that I don't understand (and don't need to).

You may want to make additional minor changes: remove the trailing `\` on lines preceding the deleted last line. It probably works as is, but relies on the presence of white lines which could break easily in the future.

golaz:
Note: this might be difficult to do for `ts.bash` due to the if statements. Maybe just leave it as is.

PR #99, opened 2021-07-14
Update Cori-Haswell ncremap command for MPAS-Analysis

xylar:
It seems that calling with `srun -n 1` hangs, whereas not having a prefix works fine.


xylar:
I tested this fix, but not through `zppy` directly.  I didn't have any luck with @czender's suggestion of adding `--gres=craynetwork:0`.  I saw the same hanging with this added.  However, possibly because of recent changes in NCO itself, I was able to run `ncremap` on Cori-Haswell without any prefix and that worked as expected.  I reran the same test that was hanging for @chengzhuzhang (and for me as well when I tested) but with this fix and it now worked.

xylar:
@forsyth2, I'm going to go ahead and merge this.  As you said on Slack, we can test further as part of `rc8`.

PR #98, opened 2021-07-14
Update E3SM-Unified activation scripts in docs

xylar:
The machine name is now included.  The location on Cori-Haswell has changed.  Cori-KNL is not recommended, so I have not included it.

PR #109, opened 2021-07-24
Add machine name to the e3sm_unified path

tangq:
The latest e3sm_unified v1.5.0 version adds machine name to its path. Updated zppy with the new paths. Otherwise, the `source e3sm_unified*.sh` command will fail.

Tested successfully on chrysalis.

### DIFF
--- a/zppy/__main__.py
+++ b/zppy/__main__.py
@@ -65,7 +65,7 @@ def main():
             )
         elif tmp.startswith("cori"):
             machine = "cori"
-            environment_commands = "source /global/cfs/cdirs/e3sm/software/anaconda_envs/load_{}_e3sm_unified_cori.sh".format(
+            environment_commands = "source /global/cfs/cdirs/e3sm/software/anaconda_envs/load_{}_e3sm_unified_cori-haswell.sh".format(
                 config["default"]["e3sm_unified"]
             )
         elif tmp.startswith("blues"):
--- a/zppy/__main__.py
+++ b/zppy/__main__.py
@@ -59,26 +59,26 @@ def main():
         if tmp.startswith("compy"):
             machine = "compy"
             environment_commands = (
-                "source /share/apps/E3SM/conda_envs/load_{}_e3sm_unified.sh".format(
+                "source /share/apps/E3SM/conda_envs/load_{}_e3sm_unified_compy.sh".format(
                     config["default"]["e3sm_unified"]
                 )
             )
         elif tmp.startswith("cori"):
             machine = "cori"
-            environment_commands = "source /global/cfs/cdirs/e3sm/software/anaconda_envs/load_{}_e3sm_unified.sh".format(
+            environment_commands = "source /global/cfs/cdirs/e3sm/software/anaconda_envs/load_{}_e3sm_unified_cori.sh".format(
                 config["default"]["e3sm_unified"]
             )
         elif tmp.startswith("blues"):
             machine = "anvil"
             environment_commands = (
-                "source /lcrc/soft/climate/e3sm-unified/load_{}_e3sm_unified.sh".format(
+                "source /lcrc/soft/climate/e3sm-unified/load_{}_e3sm_unified_anvil.sh".format(
                     config["default"]["e3sm_unified"]
                 )
             )
         elif tmp.startswith("chr"):
             machine = "chrysalis"
             environment_commands = (
-                "source /lcrc/soft/climate/e3sm-unified/load_{}_e3sm_unified.sh".format(
+                "source /lcrc/soft/climate/e3sm-unified/load_{}_e3sm_unified_chrysalis.sh".format(
                     config["default"]["e3sm_unified"]
                 )
             )


xylar:
@forsyth2, @golaz and @chengzhuzhang, is this going to make zppy in the recent release of E3SM-Unified pretty hard for users to make proper use of? If so, that would be disappointing.

tangq:
@xylar , this PR should fix it, but the users will have to reinstall zppy after the PR is merged to main.

xylar:
@tangq, I get that but that means zppy from E3SM-Unified is of limited use, doesn't it? Do I need to create a new release after this fix? :-(

tangq:
I see what you mean now, @xylar . If we want to use the zppy in e3sm_unified, it will need the old paths. I am not sure what is the best way to do that.

zppy/__main__.py
```diff
@@ -59,26 +59,26 @@ def main():
...
             environment_commands = (
-                "source /share/apps/E3SM/conda_envs/load_{}_e3sm_unified.sh".format(
+                "source /share/apps/E3SM/conda_envs/load_{}_e3sm_unified_compy.sh".format(
                     config["default"]["e3sm_unified"]
                 )
             )
         elif tmp.startswith("cori"):
             machine = "cori"
-            environment_commands = "source /global/cfs/cdirs/e3sm/software/anaconda_envs/load_{}_e3sm_unified.sh".format(
+            environment_commands = "source /global/cfs/cdirs/e3sm/software/anaconda_envs/load_{}_e3sm_unified_cori.sh".format(
```
forsyth2, line 68: This one needs to be `source /global/common/software/e3sm/anaconda_envs/load_{}_e3sm_unified_cori-haswell.sh`.

forsyth2 (changes requested):
@tangq Thanks for addressing this. I just have one comment about Cori's file path; all the paths need to match those in #98.

@xylar I remember you making the fixes in the documentation in #98 but I forgot about the default `environment_commands`. Luckily, it's quite easy for the user to manually set `environment_commands=[...]`. So, `zppy` is certainly still usable with the latest E3SM Unified, but I understand the point that it is not as seamless as it would otherwise be.

forsyth2:
@tangq I also notice the `pre-commit-hooks` check is failing. You can run `pre-commit run --all-files` before you make your commit to see what changes you need to do. If that doesn't work for you, you'll need to make sure you have the pre-commit hooks set up; see https://e3sm-project.github.io/zppy/_build/html/main/getting_started.html#b-development-environment

zppy/__main__.py
tangq: Good catch, @forsyth2 . The cori path was fixed in the new commit.

tangq:
> @tangq I also notice the `pre-commit-hooks` check is failing. You can run `pre-commit run --all-files` before you make your commit to see what changes you need to do. If that doesn't work for you, you'll need to make sure you have the pre-commit hooks set up; see https://e3sm-project.github.io/zppy/_build/html/main/getting_started.html#b-development-environment

I didn't know the pre-commit hooks. I will use them for my future PRs. It's also good to know the `environment_commands=[...]` option.

xylar:
@forsyth2, it seems important to update the documentation for the 1.0.0 release (as opposed to the hopefully separate documentation that you maintain for master) to specify that users will not be able to use the default activation script locations and *must* specify `environment_commands=...`.

forsyth2:
>  it seems important to update the documentation for the 1.0.0 release 

@tomvothecoder Is that possible? https://e3sm-project.github.io/zppy/_build/html/v1.0.0/index.html is built using the code for https://github.com/E3SM-Project/zppy/releases/tag/v1.0.0. Am I correct in assuming that there's no way to change those docs now that `v1.0.0` has been released?

rljacob:
Why is zppy sourcing those scripts internally?  The usage model for e3sm_unified is a user sources the startup script, and then they can use all the tools in e3sm_unified.

xylar:
@rljacob, `zppy` runs out of E3SM-Unified but the scripts that it generates also need to activate a conda environment. This is supposed to be E3SM-Unified by default but the appropriate update to the activation scripts got missed.

My proposal was that `zppy` should source the *same* E3SM-Unified environment (or, if zppy is not run in E3SM-Unified, activate the same conda environment in its scripts as it was run from) by default.  But I believe this was deemed too complicated for v1.0.0.  I still think it is the best default for future releases.

xylar:
>  Am I correct in assuming that there's no way to change those docs now that v1.0.0 has been released?

@forsyth2, you can make a branch off of v1.0.0, manually render the HTML locally, and make a PR to update the HTML to the `gh-pages` branch: https://github.com/E3SM-Project/zppy/tree/gh-pages/_build/html/v1.0.0.  It's definitely doable.  I've done so in my own packages.  It's not a workflow you want to do often but if the docs are not correct, there needs to be a way to fix them.

tomvothecoder:
@forsyth2 @xylar The docs are automated with GitHub Actions using `sphinx-multiversion`, which finds git tags and generates docs for them if they have a properly structured `docs/source` directory. The workflow job wipes the `gh-pages` branch and regenerates all docs whenever the job is triggered (with latest updates to `main` docs or with docs for a new git tag release).

I don't think there is an issue with how docs are automated, but instead we need to ensure our docs are correct with every release. Similarly, ReadTheDocs allows you build docs for specified tags, which means you can't fix the docs for a tag without rereleasing it.

However, if we need to fix docs for already released tags on rare occasions (like this case), we need to update the bash commands to not `rm -r *` the `gh-pages` branch and instead `rm` only the HTML build for the related branch that triggered the build (aka `main`). This means the HTML builds for tags will be preserved and you can manually update them as needed.

Related lines in `release_workflow.yml`:
https://github.com/E3SM-Project/zppy/blob/48e0b1bef1b3028ec3498923f2f6a156d6db1fc7/.github/workflows/release_workflow.yml#L91-L99

@forsyth2 I have to catch up on my work this week since I was out last week with training. It would great if you can handle this by using your fork for testing. I'll provide support where needed.

tomvothecoder:
@forsyth2 Actually, I wonder if you can just checkout the tag for `v1.0.0` and push an update to the source directly, then rebuild docs? Might want to give that a shot before messing with the CI/CD.

xylar:
@tomvothecoder, I'm afraid that won't work.  `v1.0.0` is a tag, not a branch.  You can't push to it and it would be a very bad move to update it.  It is supposed to be fixed to a specific git hash.

tomvothecoder:
@xylar Yeah I just read up on that, it is bad practice. Makes sense why you wouldn't want to do that.

tomvothecoder:
@forsyth2 I just realized when I cleaned up e3sm-diags in https://github.com/E3SM-Project/e3sm_diags/issues/306, I accidentally removed `.png` files from docs for `v2.5.0`. I'll need to do the same manual HTML build fix for e3sm-diags, so you can probably leave this for me to fix.

tomvothecoder:
@forsyth2 I'm working on updating the CI/CD build to allow manual doc updates for e3sm-diags tag releases here: https://github.com/E3SM-Project/e3sm_diags/issues/496.

You will be able to copy the commands from e3sm-diags `build_workflow.yml` and `release_workflow.yml` once it is merged.

forsyth2:
@tangq Do you want me to merge these changes? I can make the same changes but get pre-commit-hooks to pass.

tangq:
@forsyth2 , yes, please merge this PR. Thanks!

forsyth2:
Changes (along with fixes to pass pre-commit checks) merged in #118.

PR #124, opened 2021-08-20
Add extra_vars option for in the "ts" section

tangq:
This option is needed when regular expressions are used to specify variables for ncclimo. This makes the variable list more flexible.

For example, if the variables are specified with a mix of wildcards and variable names like
ncclimo -v area,'^DF_', it will fail.

But ncclimo -v '^DF_' --var_xtr=area will work.

This PR adds the --var_xtr option.

I tested it successfully on compy and chrysalis.


forsyth2 (approved):
Thanks, this change should help with the implementation of #31. 

The changes here look good to me. Be sure to choose "Squash and merge" when merging, or if you'd like, I can merge the PR.

tangq:
It is good to know that this PR helps #31 as well. I am merging it now.

PR #135, opened 2021-09-15
Add campaign parameter

### LINKED ISSUE #122: Add parameter to specify science campaign
Add parameter to specify science campaign. For example: if `campaign = "water_cycle"` then set E3SM Diags `sets` to include particular sets.

forsyth2:
Add campaign parameter. Resolves #122.

### DIFF
--- a/zppy/e3sm_diags.py
+++ b/zppy/e3sm_diags.py
@@ -26,6 +26,21 @@ def e3sm_diags(config, scriptDir):
         if "ts_num_years" in c.keys():
             c["ts_num_years"] = int(c["ts_num_years"])
 
+        if c["campaign"] == "water_cycle":
+            c["sets"] = [
+                "lat_lon",
+                "zonal_mean_xy",
+                "zonal_mean_2d",
+                "polar",
+                "cosp_histogram",
+                "meridional_mean_2d",
+                "enso_diags",
+                "qbo",
+                "diurnal_cycle",
+                "annual_cycle_zonal_mean",
+                "streamflow",
+            ]
+
         # Loop over year sets
         year_sets = getYears(c["years"])
         for s in year_sets:
--- a/zppy/mpas_analysis.py
+++ b/zppy/mpas_analysis.py
@@ -37,6 +37,27 @@ def mpas_analysis(config, scriptDir):
             c["subdir_ocean"] = "/run"
             c["subdir_ice"] = "/run"
 
+        if c["campaign"] == "water_cycle":
+            c["generate"] = [
+                "all",
+                "no_landIceCavities",
+                "no_BGC",
+                "no_icebergs",
+                "no_min",
+                "no_max",
+                "no_sose",
+                "no_climatologyMapAntarcticMelt",
+                "no_regionalTSDiagrams",
+                "no_timeSeriesAntarcticMelt",
+                "no_timeSeriesOceanRegions",
+                "no_climatologyMapSose",
+                "no_woceTransects",
+                "no_soseTransects",
+                "no_geojsonTransects",
+                "no_oceanRegionalProfiles",
+                "no_hovmollerOceanRegions",
+            ]
+
         # Loop over year sets
         ts_year_sets = getYears(c["ts_years"])
         if c["climo_years"] != [""]:
--- a/zppy/templates/default.ini
+++ b/zppy/templates/default.ini
@@ -8,6 +8,7 @@ debug = boolean(default=False)
 partition = string(default="")
 e3sm_unified = option('latest', 'test', default='latest')
 dry_run = boolean(default=False)
+campaign = string(default="custom")
 
 [climo]
 active = boolean(default=True)


forsyth2:
@xylar @wlin7 Is this what you had in mind? I specify `campaign = "water_cycle"` in the configuration file and then `zppy` will determine how to define `sets` for E3SM Diags. If both `campaign` and `sets` are specified, `campaign` will take precedence.

@golaz If this implementation looks ok, I'll also update docs/tests. The functionality is added to `zppy/e3sm_diags.py` rather than `zppy/templates/e3sm_diags.bash`.

xylar:
@forsyth2, this looks right to me.  Could you add the same functionality to the tasks from MPAS-Analysis that are included by default for the WC campaign?  Then, I can make a separate PR (and get help from @wlin7) to define the cryo configuration.

forsyth2:
> Could you add the same functionality to the tasks from MPAS-Analysis that are included by default for the WC campaign

@xylar, I'm not sure what's specific to the WC campaign for MPAS-Analysis

https://github.com/E3SM-Project/zppy/blob/main/zppy/templates/default.ini#L144 defines the default values listed below. Which parameters have WC campaign specific defaults? If I know that then I can set the defaults according to campaign. (For E3SM Diags, it looks like the only parameter that was WC-specific was `sets`).
```
active = boolean(default=True)
shortTermArchive = boolean(default=True)
qos = string(default="regular")
nodes = integer(default=1)
walltime = string(default="06:00:00")
parallelTaskCount = integer(default=12)
ncclimoParallelMode = string(default="bck")
ncclimoThreads = integer(default=12)
mapMpiTasks = integer(default=6)
cache = boolean(default=True)
purge = boolean(default=False)
PostMOC = boolean(default=False)
mpaso_nml = string(default="mpaso_in")
mpassi_nml = string(default="mpassi_in")
stream_ocn = string(default="streams.ocean")
stream_ice = string(default="streams.seaice")
anomalyRefYear = integer(default=1)
generate = string_list(default=list('all', 'no_landIceCavities', 'no_BGC', 'no_icebergs', 'no_min', 'no_max', 'no_sose', 'no_climatologyMapAntarcticMelt', 'no_regionalTSDiagrams', 'no_timeSeriesAntarcticMelt', 'no_timeSeriesOceanRegions', 'no_climatologyMapSose', 'no_woceTransects', 'no_soseTransects', 'no_geojsonTransects', 'no_oceanRegionalProfiles', 'no_hovmollerOceanRegions'))
```

xylar:
@forsyth2, thanks for your question.

The equivalent of the `set` you have for `e3sm_diags` is the `generate` string.  That should be specific to each campaign.  There would be a config file I would also add for the `cryo` campaign but I haven't figured out how to do that yet (it's not part of the MPAS-Analysis package yet.

golaz:
Hi @forsyth2 : this looks reasonable. What happens if someone uses an older cfg and ends up the default value ("custom") for the new campaign parameter? Will everything still run? Also, can a user still select "water_cycle" and override the E3SM Diags sets to something else?

forsyth2:
> What happens if someone uses an older cfg and ends up the default value ("custom") for the new campaign parameter? Will everything still run?

Good point -- I'll test it out. I believe that since there's no special directions for "custom", the default values for `sets` and `generate` would be used, as would be expected.

> can a user still select "water_cycle" and override the E3SM Diags sets to something else?

With this implementation, no. The campaign overrides the `sets`. I assume if someone wanted to change the `sets`, then they would then be using "custom" for `campaign` rather than the "water_cycle"

xylar:
@forsyth2, I wonder if, rather than"hiding" the WC config options in a python module we could expose them in a config file analogous to default.ini? This is the approach we take in MPAS-Analysis. This would make it easier for a user to quickly see all WC config options, and maybe copy and tweak them for a custom configuration.

golaz:
@xylar, that would be a good option, but can it be done? 

For Water Cycle, we would want `sets` to take on one default, and for Cryosphere to take on another one. If this can be done within `default.ini`, that would be great. I just would not know how to do it.

forsyth2:
> can it be done?

We could set `c["sets"]` to be a list generated by reading a list of sets from a file. (Similar for`c["generate"]`). However, that still doesn't give the user a good way to override certain sets from a campaign's set list.

golaz:
@forsyth2 : if @xylar has a way to accomplish this from within `default.ini`, we can do this. Otherwise your current approach would work just fine.

xylar:
My suggestion was not to include this in defaults.ini but rather to have water_cycle.ini next to defaults.ini that would override the sets option for e3sm_diags and the generate option for mpas-analysis. You would first parse defaults.ini then water_cycle.ini (then maybe a machine-specific config file). This approach works well for the normal Python config parser but maybe not for the non-standard one you are using for zppy.

xylar:
The use of a non-standard config parser here may also be an issue when interfacing with mache. We should discuss this sometime soon.

xylar:
I will be in Los Alamos starting on Tuesday.  Once I'm settled in, if you like, I can make a suggested alternative to this branch with `water_cycle.ini` and you can let me know what you think.

forsyth2:
>  I can make a suggested alternative to this branch with `water_cycle.ini`

That sounds great, thanks @xylar

xylar:
@forsyth2 and @golaz, I finally had a chance to put together my suggested alternative to this PR: https://github.com/E3SM-Project/zppy/pull/138.  Let's discuss that PR and decide what we want to do.

PR #138, opened 2021-09-27
Add campaign parameter

xylar:
This merge adds a new "campaign" parameter to the "defaults" section that is "custom" by default.  For other values (currently only "water_cycle"), a config file is loaded with alternative default parameters.  The campaign parameters take priority over the defaults but the user parameters take priority over the campaign parameters. Thus, a user could set the campaign to "water_cycle" but still modify the set of analyses to run (e.g. for debugging).

closes #135

### DIFF
--- a/docs/source/parameters.rst
+++ b/docs/source/parameters.rst
@@ -40,8 +40,8 @@ types and default values for the parameters. ::
         e3sm_unified = option('latest', 'test', default='latest')
         # This should be set to True if you don't want the batch jobs to be submitted
         dry_run = boolean(default=False)
-	# Specify which campaign you are running.
-	campaign = string(default="none")
+        # Specify which campaign you are running.
+        campaign = string(default="none")
         # The following parameter is not defined in `default.ini`
         # Set up the environment -- this is where you can tell zppy to use a custom conda environment.
         # To use a custom conda environment, you can set `environment_commands="source <path to conda.sh>; conda activate <custom environment>"`.
--- a/docs/source/parameters.rst
+++ b/docs/source/parameters.rst
@@ -40,6 +40,8 @@ types and default values for the parameters. ::
         e3sm_unified = option('latest', 'test', default='latest')
         # This should be set to True if you don't want the batch jobs to be submitted
         dry_run = boolean(default=False)
+	# Specify which campaign you are running.
+	campaign = string(default="none")
         # The following parameter is not defined in `default.ini`
         # Set up the environment -- this is where you can tell zppy to use a custom conda environment.
         # To use a custom conda environment, you can set `environment_commands="source <path to conda.sh>; conda activate <custom environment>"`.
--- a/zppy/__main__.py
+++ b/zppy/__main__.py
@@ -31,6 +31,7 @@ def main():
     # Read configuration file and validate it
     default_config = os.path.join(templateDir, "default.ini")
     user_config = ConfigObj(args.config, configspec=default_config)
+    _validate_config(user_config)
     campaign = user_config["default"]["campaign"]
     if campaign != "custom":
         campaign_file = os.path.join(templateDir, "{}.cfg".format(campaign))
@@ -39,20 +40,12 @@ def main():
                 "{} does not appear to be a known campaign".format(campaign)
             )
         config = ConfigObj(campaign_file, configspec=default_config)
+        _validate_config(config)
         # merge such that user_config takes priority
         config.merge(user_config)
     else:
         # no need to merge
         config = user_config
-
-    validator = Validator()
-
-    result = config.validate(validator)
-    if result is not True:
-        print("Validation results={}".format(result))
-        raise Exception("Configuration file validation failed")
-    else:
-        print("Configuration file validation passed")
 
     # Add templateDir to config
     config["default"]["templateDir"] = templateDir
@@ -115,3 +108,14 @@ def main():
 
     # global time series tasks
     global_time_series(config, scriptDir)
+
+
+def _validate_config(config):
+    validator = Validator()
+
+    result = config.validate(validator)
+    if result is not True:
+        print("Validation results={}".format(result))
+        raise Exception("Configuration file validation failed")
+    else:
+        print("Configuration file validation passed")
--- a/zppy/__main__.py
+++ b/zppy/__main__.py
@@ -29,7 +29,22 @@ def main():
     templateDir = os.path.join(os.path.dirname(__file__), "templates")
 
     # Read configuration file and validate it
-    config = ConfigObj(args.config, configspec=os.path.join(templateDir, "default.ini"))
+    default_config = os.path.join(templateDir, "default.ini")
+    user_config = ConfigObj(args.config, configspec=default_config)
+    campaign = user_config["default"]["campaign"]
+    if campaign != "custom":
+        campaign_file = os.path.join(templateDir, "{}.cfg".format(campaign))
+        if not os.path.exists(campaign_file):
+            raise ValueError(
+                "{} does not appear to be a known campaign".format(campaign)
+            )
+        config = ConfigObj(campaign_file, configspec=default_config)
+        # merge such that user_config takes priority
+        config.merge(user_config)
+    else:
+        # no need to merge
+        config = user_config
+
     validator = Validator()
 
     result = config.validate(validator)
--- a/zppy/__main__.py
+++ b/zppy/__main__.py
@@ -33,7 +33,7 @@ def main():
     user_config = ConfigObj(args.config, configspec=default_config)
     _validate_config(user_config)
     campaign = user_config["default"]["campaign"]
-    if campaign != "custom":
+    if campaign != "none":
         campaign_file = os.path.join(templateDir, "{}.cfg".format(campaign))
         if not os.path.exists(campaign_file):
             raise ValueError(
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ zppy.egg-info/
 
 tests/integration/image_check_failures
 test_bash_generation_output
+*.cfg.txt
 
 # Sphinx documentation
 docs/_build/
--- a/docs/source/post.mysimulation.cfg
+++ b/docs/source/post.mysimulation.cfg
@@ -6,6 +6,7 @@ case = 20210122.v2_test01.piControl.ne30pg2_EC30to60E2r2-1900_ICG.chrysalis
 www = <www>
 e3sm_unified = latest
 partition = compute
+campaign = "water_cycle"
 
 # Regridded atmosphere climatologies every 20 and 50 years
 [climo]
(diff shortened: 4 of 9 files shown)


xylar:
@forsyth2, could you give me an example config file I can use as a starting point to test this quickly on Chrysalis?  So far, I haven't tested it at all.

xylar:
I believe the build fails are expected because we added a new config option.

forsyth2:
Thanks @xylar!

> could you give me an example config file

This is the config file I was testing with:
```
[default]
input = /lcrc/group/e3sm/ac.forsyth2/E3SMv2/v2.LR.historical_0201
output = /lcrc/group/e3sm/ac.forsyth2/zppy_development/issue_122.v2.LR.historical_0201
case = v2.LR.historical_0201
www = /lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2/zppy_development/issue_122/
partition = debug
environment_commands = "source /lcrc/soft/climate/e3sm-unified/load_latest_e3sm_unified_chrysalis.sh"
campaign = "water_cycle"

[climo]
active = True
years = "1985-2014",

  [[ atm_monthly_180x360_aave ]]
  input_subdir = "archive/atm/hist"
  mapping_file = /home/ac.zender/data/maps/map_ne30pg2_to_cmip6_180x360_aave.20200201.nc
  frequency = "monthly"

  [[ atm_monthly_diurnal_8xdaily_180x360_aave ]]
  input_subdir = "archive/atm/hist"
  input_files = "eam.h4"
  mapping_file = /home/ac.zender/data/maps/map_ne30pg2_to_cmip6_180x360_aave.20200201.nc
  vars = "PRECT"
  frequency = "monthly_diurnal_8xdaily"

[ts]
active = True
years = "1985:2014:10",

  [[ atm_monthly_180x360_aave ]]
  input_subdir = "archive/atm/hist"
  input_files = "eam.h0"
  frequency = "monthly"
  mapping_file = /home/ac.zender/data/maps/map_ne30pg2_to_cmip6_180x360_aave.20200201.nc
  # Use default vars

  [[ atm_monthly_glb ]]
  input_subdir = "archive/atm/hist"
  input_files = "eam.h0"
  frequency = "monthly"
  mapping_file = "glb"
  # Use default vars

  [[ rof_monthly ]]
  input_subdir = "archive/rof/hist"
  input_files = "mosart.h0"
  frequency = "monthly"
  mapping_file = ""
  vars = "RIVER_DISCHARGE_OVER_LAND_LIQ"
  extra_vars = 'areatotal2'

[e3sm_diags]
active = True
years = "1985-2014",
ts_num_years = 10
ref_start_yr = 1985
ref_final_yr = 2014
environment_commands="source /home/ac.forsyth2/miniconda3/etc/profile.d/conda.sh; conda activate e3sm_diags_env_dev_20210827"

  [[ atm_monthly_180x360_aave ]]
  short_name = 'v2.LR.historical_0201'
  grid = '180x360_aave'
  reference_data_path = '/lcrc/soft/climate/e3sm_diags_data/obs_for_e3sm_diags/climatology'
  obs_ts = '/lcrc/soft/climate/e3sm_diags_data/obs_for_e3sm_diags/time-series'
  dc_obs_climo = '/lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/obs/climatology'
  climo_diurnal_subsection = "atm_monthly_diurnal_8xdaily_180x360_aave"
  output_format_subplot = "pdf",
  streamflow_obs_ts = '/lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/obs/time-series/'

[e3sm_diags_vs_model]
active = False

[amwg]
active = False

[mpas_analysis]
active = True
partition = compute
walltime = "6:00:00"
parallelTaskCount = 6
anomalyRefYear = 1985
ts_years = "1985-2014",
enso_years = "1985-2014",
climo_years = "1985-2014",
mesh = "EC30to60E2r2"

[global_time_series]
active = False
years = "1985-2014",
ts_num_years = 10
figstr= "v2_historical_0201"
moc_file = "mocTimeSeries_1985-2014.nc"
experiment_name = "v2.LR.historical_0201"
ts_years = "1985-2014",
climo_years = "1985-2014",
```

> I believe the build fails are expected because we added a new config option.

Yes, that's correct. I wasn't going to worry about updating the tests until we settled on an implementation.

forsyth2:
@xylar I tested out your code and it works fine with `campaign = "water_cycle"`. However, it looks like `campaign` has to be specified explicitly for it to run -- this might be ok though. Other than that, this implementation looks good to me.

xylar:
## Testing

I tested on Chrysalis using a slightly modified version of @forsyth2's config file above.  

First, I tested with the `water_cycle.cfg` from this branch as it is (which is the same as `default.ini`):
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/zppy_dev/campaign-config-file/v2.LR.historical_0201/

Then, I modified `water_cycle.cfg` to just run a few analysis sets in both e3sm_diags and mpas-analysis.  I wanted to make sure those changes got picked up, and they did:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/zppy_dev/campaign-config-file-wc-mod/v2.LR.historical_0201/

Finally, I reset `water_cycle.cfg` back to the version in this branch and instead modified the set of analysis to run my user config file.  I wanted to make sure that the user config file took priority.  This, too, worked as expected:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/zppy_dev/campaign-config-file-user-mod/v2.LR.historical_0201/

xylar:
> However, it looks like campaign has to be specified explicitly for it to run -- this might be ok though.

Hmm, that wasn't the intention.  It should get `campaign = custom` from `default.ini`.  Let me try that out.

xylar:
@forsyth2, I'm seeing:
```
Traceback (most recent call last):
  File "/home/ac.xylar/chrysalis/miniconda3/envs/zppy_dev/bin/zppy", line 8, in <module>
    sys.exit(main())
  File "/home/ac.xylar/chrysalis/miniconda3/envs/zppy_dev/lib/python3.7/site-packages/zppy/__main__.py", line 34, in main
    campaign = user_config["default"]["campaign"]
  File "/home/ac.xylar/chrysalis/miniconda3/envs/zppy_dev/lib/python3.7/site-packages/configobj.py", line 554, in __getitem__
    val = dict.__getitem__(self, key)
KeyError: 'campaign'
```
when I don't have `campaign` in my user config file.  Presumably that's what you saw, too.  But shouldn't there be a default value from `default.ini`?
```
campaign = string(default="custom")
```
Do you understand why that's not the case?

xylar:
I think the answer may be that you need to run the validator?

xylar:
Yes, that seems to be it.

forsyth2:
> Presumably that's what you saw, too. 

Yes, that's the error I got.

xylar:
@forsyth2, see if it works for you with my latest changes.

forsyth2:
@xylar Thanks! I think this approach is good. Editing `zppy/templates/<campaign>.cfg` is a cleaner approach than editing the zppy/<task>.py` files directly. 

Do you want me to take over from here to update tests, documentation, etc. or do you want to do that?

xylar:
> Editing `zppy/templates/<campaign>.cfg` is a cleaner approach than editing the `zppy/<task>.py` files directly.

I would expect that users would not edit either of these files, but I guess you mean for developers?  To me, the more important point is for users.  First, if a user changes a parameter in a user config file, that will not be overridden by the campaign config file in this implementation.  Also, a user can look at `zppy/templates/<campaign>.cfg` and copy the config options into the user config file to make changes they might need, whereas they would never think to look in the python files for this.

> Do you want me to take over from here to update tests, documentation, etc. or do you want to do that?

Yes, go for it!

forsyth2:
Yes, I meant the approach is cleaner for development, but it certainly benefits users as well!

zppy/__main__.py
```diff
@@ -29,15 +29,23 @@ def main():
...
-
-    result = config.validate(validator)
-    if result is not True:
-        print("Validation results={}".format(result))
-        raise Exception("Configuration file validation failed")
+    default_config = os.path.join(templateDir, "default.ini")
+    user_config = ConfigObj(args.config, configspec=default_config)
+    _validate_config(user_config)
+    campaign = user_config["default"]["campaign"]
+    if campaign != "custom":
```
forsyth2, line 36: Perhaps a better default `campaign` would be `"default_settings"`, since this setting simply tells `zppy` to use the `default.ini` values for `sets` and `generate`.

zppy/__main__.py
```diff
@@ -29,15 +29,23 @@ def main():
...
+        campaign_file = os.path.join(templateDir, "{}.cfg".format(campaign))
```
forsyth2, line 37: If the user wants to create their own config file, does it also have to be in `templateDir` then -- i.e., ``zppy/templates/<my-custom-campaign>.cfg``?

zppy/__main__.py
xylar: I think I would find `campaign = "none"` to be a more intuitive default.  Since the defaults are now (and probably will continue to be) the same as the `water_cycle`, we could just make `water_cycle` the default.

zppy/__main__.py
xylar: No, the user puts things into the usual config file that's passed to `zppy` with `-c`.  There is no added advantage in putting a file in `templateDir` unless it's something to be shared among everyone, like the campaign defaults.

xylar:
@forsyth2, I'm going to rename "custom" to "none".  I was just following your lead from #135.

xylar:
@forsyth2, I just invited you to be a collaborator on my fork in case you want to push changes to this branch.

forsyth2:
@xylar Thanks, I added a commit for testing and documentation. Let me know if it looks good to you. Then, we should be good to merge with "Squash and merge".

docs/source/parameters.rst
```diff
@@ -40,6 +40,8 @@ types and default values for the parameters. ::
         e3sm_unified = option('latest', 'test', default='latest')
         # This should be set to True if you don't want the batch jobs to be submitted
         dry_run = boolean(default=False)
+	# Specify which campaign you are running.
+	campaign = string(default="none")
```
xylar, line 44: These are tabs but should be spaces

xylar:
@forsyth2, just a tiny formatting issue.  Then, go for it!

xylar:
Looks great!  I'll add support for `campaign = "cryosphere"` and `cryosphere.cfg` in a separate PR once this goes in and once https://github.com/MPAS-Dev/MPAS-Analysis/pull/827 has gone in to support the cryosphere-specific config options for MPAS-Analysis.

xylar:
Thanks @forsyth2

forsyth2:
Thanks for implementing the configuration file idea!

golaz:
@xylar : thank you for implementing this very elegant solution. I really like your idea of having separate (but non-redundant) cfg files for different simulation campaigns.

PR #143, opened 2021-10-06
Fix validation of campaign configs

xylar:
This merge switches to first merging user and campaign config files and only validating after the merge.  This slightly complicates the logic for getting the campaign (if any) because defaults have not yet been parsed at that point.  But it does mean we only validate once (so only one validation success is printed) and it means that defaults will not override the campaign config options, as was happening before this merge.

### DIFF
--- a/docs/source/dev_guide/testing.rst
+++ b/docs/source/dev_guide/testing.rst
@@ -64,6 +64,25 @@ To replace the expected bash files for ``test_bash_generation.py`` run the follo
        # Rerun test
        python -m unittest tests/integration/test_bash_generation.py       
 
+To replace the expected files for ``test_campaign.py`` run the following:
+
+    .. code::
+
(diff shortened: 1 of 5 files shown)


xylar:
I'm not sure how this was missed in my testing of #138.

xylar:
@forsyth2, let me know if you want me to include that change.

forsyth2:
> @forsyth2, let me know if you want me to include that change.

@xylar Yeah, if you don't mind, that would be great. Thanks! I personally like seeing the "validation passed" statement, but yes, probably most users won't care as long as it works.

xylar:
@forsyth2, thanks very much for testing.  I think you can merge this whenever you're ready.

forsyth2:
The commit I would add to this branch is here: https://github.com/forsyth2/zppy/commit/245b735e46052b94ef718c5286241624aae00837

xylar:
@forsyth2, that looks good to me.

PR #144, opened 2021-10-06
Add support for cryosphere campaign

xylar:
This includes a new cryosphere config file in the templates and a special `--polar_regions` flag for MPAS-Analysis.  The new flag will only be supported for MPAS-Analysis >= 1.5.0 (not yet released), so it will require an update of E3SM-Unified to work correctly.

### DIFF
--- a/docs/source/dev_guide/testing.rst
+++ b/docs/source/dev_guide/testing.rst
@@ -68,18 +68,24 @@ To replace the expected files for ``test_campaign.py`` run the following:
 
     .. code::
 
+       rm -rf /lcrc/group/e3sm/public_html/zppy_test_resources/test_campaign_cryosphere_expected_files
+       rm -rf /lcrc/group/e3sm/public_html/zppy_test_resources/test_campaign_cryosphere_override_expected_files
        rm -rf /lcrc/group/e3sm/public_html/zppy_test_resources/test_campaign_none_expected_files 
        rm -rf /lcrc/group/e3sm/public_html/zppy_test_resources/test_campaign_water_cycle_expected_files
        rm -rf /lcrc/group/e3sm/public_html/zppy_test_resources/test_campaign_water_cycle_override_expected_files
+       mkdir -p /lcrc/group/e3sm/public_html/zppy_test_resources/test_campaign_cryosphere_expected_files
+       mkdir -p /lcrc/group/e3sm/public_html/zppy_test_resources/test_campaign_cryosphere_override_expected_files
        mkdir -p /lcrc/group/e3sm/public_html/zppy_test_resources/test_campaign_none_expected_files
        mkdir -p /lcrc/group/e3sm/public_html/zppy_test_resources/test_campaign_water_cycle_expected_files
        mkdir -p /lcrc/group/e3sm/public_html/zppy_test_resources/test_campaign_water_cycle_override_expected_files
        cd <top level of zppy repo>
        # Your output will now become the new expectation.
        # You can just move (i.e., not copy) the output since re-running this test will re-generate the output.
-       mv test_campaign_none_output/post/scripts /lcrc/group/e3sm/public_html/zppy_test_resources/test_campaign_none_expected_files 
-       mv test_campaign_water_cycle_output/post/scripts /lcrc/group/e3sm/public_html/zppy_test_resources/test_campaign_water_cycle_expected_files
-       mv test_campaign_water_cycle_override_output/post/scripts /lcrc/group/e3sm/public_html/zppy_test_resources/test_campaign_water_cycle_override_expected_files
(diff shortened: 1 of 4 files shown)


zppy/templates/cryosphere.cfg
```diff
@@ -0,0 +1,8 @@
+[e3sm_diags]
+sets = "lat_lon","zonal_mean_xy","zonal_mean_2d","polar","cosp_histogram","meridional_mean_2d","enso_diags","qbo","diurnal_cycle","annual_cycle_zonal_mean","streamflow"
+
+[e3sm_diags_vs_model]
+sets = "lat_lon","zonal_mean_xy","zonal_mean_2d","polar","cosp_histogram","meridional_mean_2d"
+
```
xylar, line 6: @wlin7, could you take a look at this list or flag someone to take a look?  I would like to make sure we're running the analysis that you would want for cryosphere runs.

If there are other zppy config options you would like to change from the defaults for cryosphere runs, here is the place to add them.

If you would like to change other things that zppy doesn't yet support, this PR would be a good place to make those changes, too.  (As long as they're not too complicated.)

zppy/templates/cryosphere.cfg
```diff
@@ -0,0 +1,8 @@
...
+[mpas_analysis]
+generate = 'all', 'no_BGC', 'no_icebergs'
```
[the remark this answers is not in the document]
stephenprice: I seem to have the following in a few of my configs as well:

'no_index', 'no_eke', 'no_min', 'no_max'

But I will confess that this is just from re-using config files that have worked in the past as opposed to having a good understanding for why these are switched off.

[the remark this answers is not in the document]
stephenprice: Agreed.

xylar:
I just rebased.

forsyth2:
@xylar Once this is ready to merge, we can add a commit I worked on for testing (https://github.com/forsyth2/zppy/commit/d47407375e1a2b9bde664f3c6926a128d9266eac)

zppy/mpas_analysis.py
```diff
@@ -21,6 +21,8 @@ def mpas_analysis(config, scriptDir):
     if len(tasks) == 0:
         return
 
+    campaign = config["default"]["campaign"]
```
forsyth2, line 24: The changes in this file are unnecessary. I confirmed by running the integration tests with these changes removed. I also checked that `extra_config="--polar_regions"` for the cryosphere case.

forsyth2 (approved):
From the standpoint of integrating into `zppy`, this looks good to me. (I can't say what is necessary for the cryosphere campaign in particular though).

zppy/mpas_analysis.py
forsyth2: That's correct. For example, in `zppy/templates/mpas_analysis.bash`:
-  the variable ``{{ scriptDir }}`` comes from `c["scriptDir"] = scriptDir` in `zppy/mpas_analysis.py` (and in fact is not defined in `zppy/templates/default.ini` at all).
- the variable ``{{ debug }}``, on the other hand, is not defined in `main/zppy/mpas_analysis.py`` but rather only in ``zppy/templates/default.ini``. (This is the same case as `{{ campaign }}`).

Also worth noting is that the variables doesn't even have to be specified anywhere other than the bash file. For example, in `zppy/templates/e3sm_diags.bash`, ``{{ ref_start_yr }}`` isn't defined in `zppy/e3sm_diags.py` or in `zppy/templates/default.ini`. However, the user should specify that value in their config file if they are running the ENSO or QBO sets.

xylar: Thanks.  I'll take these unnecessary parts out as soon as I get some feedback on the `e3sm_diags` set from @wlin7.

zppy/templates/cryosphere.cfg
wlin7: Thanks for creating this template, @xylar. The sets for e3sm_iags vs observations are the typical onesI haven't seen what are included in 'streamflow'.

For e3sm_diags_vs_model, I would like to see 'enso_diags', and 'annual_cycle_zonal_mean' also included, for easy comparison of model differences. These could be more useful when comparing cryo simulations against WC simulations.

xylar: Thanks @wlin7.  I'll make these changes.

forsyth2:
@xylar We haven't really been using/testing `e3sm_diags_vs_model` (`active = False` in all test cases). Almost all the E3SM Diags functionality improvements have gone into the `e3sm_diags` task exclusively. Ultimately, the model-vs-model functionality will be moved into that task too (#147).

Issue #132, opened 2021-09-15
Improve output directory structure


forsyth2:
Improve output directory structure

Currently:
- atm/
  - 180x360_aave/
    - clim/
      - 100yr/
      - 250yr/
      - 50yr/
    - clim_atm_monthly_diurnal_8xdaily_180x360_aave/
      - 100yr/
      - 250yr/
      - 50yr/
    - ts/
      - daily/
      - monthly/
  - glb/
    - ts/
      - monthly/

golaz:
The problem is that the sub-directory name `clim_atm_monthly_diurnal_8xdaily_180x360_aave` duplicates unnecessary information (`atm`, `180x360_aave`).

One possible alternative would be

- atm/
  - 180x360_aave/
     -  clim/
        - monthly/   
           - 100yr/
           - 250yr/
            - 50yr/
         - monthly_diurnal_8xdaily/
            - 100yr/
            -  250yr/
            -  50yr/
     -  ts/
        -  daily/
        -  monthly/
     -  glb/
        -  ts/
           -  monthly/

But strictly speaking, the files under `clim/monthly` are not all monthly since ncclimo also generates seasonal and annual files. But maybe that's ok.

chengzhuzhang:
For TC tracking files, we could have a directory the atm/tc.
Also need a atm/arm-sites (optional) directory for high-frequency ARM-site output (but the name list variables are not available as default)
For cmorized files, it could be atm/180x360_aave/ts_cmip with sub directories /6hrly /daily /monthly. Same for other component lnd, ocean, sea-ice... Because regular CMORizing process re-org files in a CMIP specific directory system, we could have a separate system for CMORized files, which is another possibility

forsyth2:
@golaz @chengzhuzhang I'm thinking the second alternative might be best. If you both agree with that, I can start changing the directory structure.

**Second alternative**

- atm/
  - 180x360_aave/
     -  clim/
         - 100yr/
         - 250yr/
         - 50yr/
       - clim_diurnal_8xdaily/
          - 100yr/
          -  250yr/
          -  50yr/
     -  ts/
        -  daily/
        -  monthly/
   -  glb/
      -  ts/
         -  monthly/
   - tc/

golaz:
@forsyth2 : I like this second alternative as well.

Let's use the `frequency` parameter to determine the destination directory. This will decouple the sub-section name from the directory structure and thus should be more robust.

PR #169, opened 2021-11-13
Add tropical cyclone diags

### LINKED ISSUE #61: Add tropical cyclone analysis diagnostics
Add tropical cyclone analysis diagnostics now that https://github.com/E3SM-Project/e3sm_diags/pull/409 has merged.

forsyth2:
Add tropical cyclone diags. Resolves #61.

- [x] Run integration tests
- [ ] Final review & merge
- [ ] Updated expected files for integration tests


forsyth2:
Note: #132 should be incorporated into this.

zppy/templates/postprocess_tc.bash
```diff
@@ -0,0 +1,101 @@
...
+source /global/common/software/e3sm/anaconda_envs/load_latest_e3sm_unified_cori-haswell.sh
+
+#For typical EAM v2 ne gp2 grids.
+start="{{ '%04d' % (year1) }}"
+end="{{ '%04d' % (year2) }}"
+caseid="{{ case }}"
+drc_in={{ input }}/{{ input_subdir }}
+# Warning: tempest-remap can only write grid file on SCRATCH space.
+# The result files will be moved to another path at the end.
+result_dir_fin={{ output }}/post/atm/tc # Directory will be {{ output }}/post/atm/tc/tc-analysis
```
forsyth2, line 34: Is `{{ output }}/post/atm/tc/tc-analysis` a good directory path for the output?

forsyth2 (commented):
@chengzhuzhang @golaz This is ready for review. I ran it successfully and I also ran the tests (I will update expected files after merging). I think this can be merged independently of the directory restructuring I discussed in my email (for #132).

zppy/templates/postprocess_tc.bash
chengzhuzhang: Hmmm, maybe we can use `{{ output }}/post/atm/tc-analysis` directly? It seems like we won't have another sub-folder under /tc .

zppy/templates/postprocess_tc.bash
```diff
@@ -0,0 +1,101 @@
...
+mkdir -p $result_dir_fin
+result_dir={{ scratch }}/tc-analysis/
+
+res=30               # res = 30/120 for ne30/ne120 grids
+pg2=true             # Set false for v1 production simulations
+atm_name="eam"       # Use "cam" for v1 production simulations
```
chengzhuzhang, line 40: model name can be passed in from user config?

zppy/templates/postprocess_tc.bash
```diff
@@ -0,0 +1,101 @@
...
+end="{{ '%04d' % (year2) }}"
+caseid="{{ case }}"
+drc_in={{ input }}/{{ input_subdir }}
+# Warning: tempest-remap can only write grid file on SCRATCH space.
+# The result files will be moved to another path at the end.
+result_dir_fin={{ output }}/post/atm/tc # Directory will be {{ output }}/post/atm/tc/tc-analysis
+mkdir -p $result_dir_fin
+result_dir={{ scratch }}/tc-analysis/
+
+res=30               # res = 30/120 for ne30/ne120 grids
```
chengzhuzhang, line 38: Line 38 and 39 specified for ne30 pg2 grids. If both were specified in user config?

chengzhuzhang (commented):
The PR looks good overall. The only glitch I found was that, mesh generation depends on the grid type (ne30np4 vs ne30pg2). By specifying Line 38-40, it would only work for ne30pg2 type of grid with model name "eam" in file name. I know there is a model name parameter in zppy that can pass in, but is there a parameter for grid type that user can specify?

golaz:
@forsyth2; a few initial comments

- I'm not a big fan of the name [postprocess_tc], could we just simply make it [tc]?
- Could we make the task smart and automatically detect the input grid? Chrysalis is down right now, so I cannot check if it is somewhere in the netCDF metadata.

zppy/templates/postprocess_tc.bash
forsyth2: That sounds good to me

forsyth2:
@chengzhuzhang 
> is there a parameter for grid type that user can specify?

Yes, there is a `grid` parameter, so we could use that to specify ne30np4 vs ne30pg2.

@golaz 
> could we just simply make it [tc]?

I feel like it should be a little bit more descriptive, maybe something like `[generate_tc_files]`?

> Could we make the task smart and automatically detect the input grid?

Possibly. I can look into that when Chrysalis is back up.

forsyth2:
@chengzhuzhang Would this work?

In the configuration file:
```
grid = "ne30pg2"
input_files = "eam.h2"
```

In `zppy/postprocess_tc.py`:
```
            if c["grid"]:
                m = re.match("ne(\d\d)(.*)", c["grid"])
                if m:
                    c["res"] = m.group(1)
                    if "pg2" ==	m.group(2):
                        c["pg2"] = "true"
                    else:
                        c["pg2"] = "false"
                else:
                    raise Exception("grid={} is not valid.".format(c["grid"]))
            else:
                raise Exception("No value was given for `grid`.")
            if c["input_files"]:
                c["atm_name"] = c["input_files"].split(".")[0]
            else:
                raise Exception("No value was given for `input_files`.")
```
In `zppy/templates/postprocess_tc.bash`:
```
res={{ res }}
pg2={{ pg2 }}
atm_name={{ atm_name }}
```

@golaz 

> Could we make the task smart and automatically detect the input grid?

I don't think so; I only get one result looking for `ne30` in the `eam.h2` files:
```
ncdump -h v2.LR.historical_0201.eam.h2.2014-12-17-00000.nc | grep ne30
		:topography_file = "/lcrc/group/e3sm/data/inputdata/atm/cam/topo/USGS-gtopo30_ne30np4pg2_16xdel2.c20200108.nc" ;
```

chengzhuzhang:
Hey @forsyth2, to auto detect grid, i think you found the right info:
`		:topography_file = "/lcrc/group/e3sm/data/inputdata/atm/cam/topo/USGS-gtopo30_ne30np4pg2_16xdel2.c20200108.nc" ;`,
where `ne30np4pg2` is a longer version of `ne30gp2`, this topography_file attribute should exist in meta data of every history files and is valid to be used to detect grid type.
(more topography_files available here: https://web.lcrc.anl.gov/public/e3sm/inputdata/atm/cam/topo/

Once the grids are auto detected, there is no need for user-input for an additional grids parameter. And the rest of the changes look fine to me..

forsyth2:
> to auto detect grid, i think you found the right info:

@chengzhuzhang What's the best way to extract this info? I can get it by having Python look for `topography_file` in the output of `ncdump -h`, but that isn't a particularly clean solution. I tried `ncdump -v topography_file` but `topography_file` isn't a variable.

> I'm not a big fan of the name [postprocess_tc], could we just simply make it [tc]?

@golaz: `[tc]` would be fine, but I'm concerned it's too similar to `[ts]` for time-series. I'd also like a little more description of what's going on -- e.g. `[generate_tc_files]`

chengzhuzhang:
Hey Ryan, I got the following tip from @czender to filter attributes using nco command. Try this:

`ncks --trd -M -m inputfile.nc | grep -E -i "^global attribute [0-9]+: topography_file" | cut -f 11- -d ' ' | sort`
(more details can be found around the use of `function ncattget` here http://nco.sourceforge.net/nco.html)

forsyth2:
@chengzhuzhang Thanks, the following works:
```
$ ncks --trd -M -m v2.LR.historical_0201.eam.h2.2014-12-17-00000.nc | grep -E -i "^global attribute [0-9]+: topography_file" | cut -f 11- -d ' ' | sort
/lcrc/group/e3sm/data/inputdata/atm/cam/topo/USGS-gtopo30_ne30np4pg2_16xdel2.c20200108.nc
```
However, I'm still unclear how to extract the resolution/grid. Looking at the list of topography files (https://web.lcrc.anl.gov/public/e3sm/inputdata/atm/cam/topo/), not all of them have a grid of the form `ne#np#pg#`. It looks like the resolution is always the part after the first underscore though? That could work.

chengzhuzhang:
> 
>  It looks like the resolution is always the part after the first underscore though? That could work.

This this also my understanding.

golaz (approved):
Approving based on visual inspection. I will run it during testing of the next E3SM Unified.

I'm still not a big fan of the name. [postprocess_tc] is a bit odd, because it is post-processing relative to the model, but pre-processing relative to E3SM Diags. Another alternative would be [tc-analysis] to establish a clear connection between the zppy task and the output sub-directory.

chengzhuzhang (approved):
Hey Ryan, the PR looks good to me based on visual check.

zppy/templates/tc_analysis.bash
```diff
@@ -0,0 +1,118 @@
...
+GenerateCSMesh --res $res --alt --file ${result_dir}outCSne$res.g
+out_type="CGLL"
+# For v2 production simulation with pg2 grids:
+if $pg2; then
+    GenerateVolumetricMesh --in ${result_dir}outCSne$res.g  --out ${result_dir}outCSne$res.g --np 2 --uniform
+    out_type="FV"
+fi
+echo $out_type
+# Generate connectivity files (.dat)
+GenerateConnectivityFile --in_mesh ${result_dir}outCSne$res.g --out_type $out_type --out_connect ${result_dir}connect_CSne${res}_v2.dat
```
forsyth2, line 72: @chengzhuzhang This command fails with `GenerateConnectivityFile: error while loading shared libraries: libnetcdf.so.11: cannot open shared object file: No such file or directory` if two year sets are run simultaneously (e.g., `years = "1985-1990", "1985-1994",`). If I run the bash files for each year set separately, I don't get this error. I thought changing `result_dir` from `tc-analysis`  to `tc-analysis_${Y1}_${Y2}` would fix this, but apparently not.

Do you have an idea of what's going on here?

chengzhuzhang: Hey Ryan, is this an issue that, two concurrent `GenerateConnectivityFile` calls from `e3sm_unified` would cause this error? Even if they are operate on files in separate folders? I'm pinging @paullric to take a look at this error message..

forsyth2: I suppose that's possible, but I don't see why two concurrent calls would cause an issue.

zppy/tc_analysis.py
```diff
@@ -64,3 +69,8 @@ def tc_analysis(config, scriptDir):
                     # Update status file
                     with open(statusFile, "w") as f:
                         f.write("WAITING %d\n" % (jobid))
+
+                # Note that this line should still be executed even if jobid == -1
+                # The later tc_analysis tasks still depend on this task (and thus will also fail).
+                # Add to the dependency list
+                dependencies.append(statusFile)
```
forsyth2, line 76: @chengzhuzhang I realized we can do what `mpas_analysis` does to force `tc_analysis` to run sequentially. I'm testing this now, but I think this should at least provide a work-around. If the tests work, I will squash and merge this PR.

I've documented the shared library error in #180, as it would be good to run multiple `year_sets` in parallel.

chengzhuzhang: Sounds good. I think it's a reasonable temporal workaround.

PR #174, opened 2022-01-13
Fix E3SM-Unified path for Cori-Haswell and -KNL

xylar:
This merge updates the path where E3SM-Unified is found on Cori.  It also changes default behavior for loading the environment on Cori to select the appropriate environment depending on the requested partition (`haswell` or `knl`).


forsyth2 (approved):
The code change looks good to me. Thanks!

Issue #152, opened 2021-10-15
Add high resolution support for remaining tasks


forsyth2:
#136 added high resolution support for `climo` and `ts`. The other `zppy` tasks still need high resolution support.

xylar:
It might be worth considering a config option similar to `campaign` to distinguish between "low", "medium" and "high" resolution. (I put these in quotes because the distinction is subjective.)

MPAS-Analysis run on high-res data provided by @ndkeen required:
- [ ] `PostMOC = True` because high res run was conducted with an old version of the MOC online analysis that does not use the regions or dimension names required by MPAS-Analysis
- [ ] Turning off all analysis involving anomalies because the reference year isn't available:
```
generate = 'all', 'no_landIceCavities', 'no_BGC', 'no_icebergs', 'no_min', 'no_max', 'no_sose', 'no_climatologyMapAntarcticMelt', 'no_regionalTSDiagrams', 'no_timeSeriesAntarcticMelt', 'no_timeSeriesOceanRegions', 'no_climatologyMapSose', 'no_woceTransects', 'no_soseTransects', 'no_geojsonTransects', 'no_oceanRegionalProfiles', 'no_hovmollerOceanRegions',  'no_climatologyMapOHCAnomaly', 'no_timeSeriesOHCAnomaly', 'no_timeSeriesSSHAnomaly', 'no_timeSeriesSalinityAnomaly', 'no_timeSeriesTemperatureAnomaly', 'no_refYearMpasClimatologyOcean'
```
I think this may be turning off more analysis than it's leaving on so we probably want to switch to explicitly specifying the analysis to run.
- [ ] Setting `parallelTaskCount = 1` and `ncclimoThreads = 1` to prevent OOM errors. 

@ndkeen, were there other changes that I'm forgetting?

ndkeen:
I think those were the changes -- but still trying to complete the analysis.

xylar:
I was able to complete 10 years (climatologies and time series) of high-res MPAS-Analysis on the large memory nodes on Cori in well under an hour.  It seems like that might be the main plausible way of doing MPAS-Analysis at high res.

xylar:
I don't know exactly when but I will try to add this support sometime in the next week or 2.

forsyth2:
> I don't know exactly when but I will try to add this support sometime in the next week or 2.

Thanks @xylar!

xylar:
@forsyth2, I will try to make a PR for this tomorrow.

[...]

PR #176, opened 2022-01-13
Add a template for high-res E3SM v1 runs 

xylar:
Currently, all config options are for MPAS-Analysis.

### DIFF
--- a/zppy/templates/high_res_v1.cfg
+++ b/zppy/templates/high_res_v1.cfg
@@ -1,0 +1,13 @@
+[default]
+# make sure we load the Cori-Haswell environment
+partition = "haswell"
+
+[mpas_analysis]
+# Use the big-memory nodes on the amd parition (these require special access)
+qos = "bigmem"
+partition = "amd"
+parallelTaskCount = 16
+mapMpiTasks = 32
+generate = 'climatologyMapMLD', 'climatologyMapSST', 'climatologyMapSSS', 'climatologyMapSSH', 'climatologyMapEKE', 'climatologyMapWoa', 'climatologyMapArgoTemperature', 'climatologyMapArgoSalinity', 'woceTransects', 'timeSeriesSST', 'timeSeriesTransport', 'streamfunctionMOC', 'no_refYearMpasClimatologyOcean', 'climatologyMapSeaIceConcNH', 'climatologyMapSeaIceThickNH', 'climatologyMapSeaIceConcSH', 'climatologyMapSeaIceThickSH'
+mesh = "oRRS18to6v3"
+PostMOC = True


xylar:
## Testing

I am running tests with both E3SM-Unified 1.5.1 and 1.6.0rc3 on Cori.

The config file for the latter looks like this:
``` cfg
[default]
input = /global/cfs/cdirs/e3smdata/simulations/20190910.branch_noCNT.n825def.unc06.A_WCYCL1950S_CMIP6_HR.ne120_oRRS18v3_ICG
output = /global/cscratch1/sd/xylar/test_noel_high_res_zppy/e3smu_1.6.0rc3
case = 20190910.branch_noCNT.n825def.unc06.A_WCYCL1950S_CMIP6_HR.ne120_oRRS18v3_ICG
www = /global/cfs/cdirs/e3sm/www/xylar/test_noel_high_res_zppy/e3smu_1.6.0rc3
campaign = "high_res_v1"
environment_commands = "source /global/common/software/e3sm/anaconda_envs/test_e3sm_unified_1.6.0rc3_cori-haswell.sh"

[climo]
active = False
[ts]
active = False
[e3sm_diags]
active = False
[e3sm_diags_vs_model]
active = False
[amwg]
active = False

# MPAS-Analysis
[mpas_analysis]
active = True
walltime = "04:00:00"
ts_years = "56-65",
enso_years = "56-65",
climo_years = "56-65",

[global_time_series]
active = False
```

The command to run is:
```
module load cmem
zppy -c high_res.cfg
```

xylar:
@forsyth2, this is in draft mode until #174 is merged and I can rebase.

It will be important to document that the high res analysis will only work if someone has requested access to the `bigmem` nodes on Cori (https://docs.nersc.gov/systems/cori-largemem/#access-to-the-large-memory-nodes) and that `zppy` will only work if you first load the necessary module `module load cmem`.

xylar:
I'm not 100% sure this should close #152.  There may be some additional analysis that needs attention and/or should be added to the `zppy/templates/high_res_v1.cfg` file.

xylar:
@forsyth2, feel free to merge this when you're ready.  You can take out the `closes #152` message if you don't want that issue closed yet.  I will not be available for follow-up until tomorrow.

forsyth2:
@xylar Thanks for the PR. We can leave #152 open for now.

The `cfg` file looks good to me. I'll trust your testing on Cori.

I added an integration test and documentation in this commit: https://github.com/forsyth2/zppy/commit/47a3561f9849ab5ecf478873787f10314d1fb5cd -- let me know if you'd like me to include that in this PR or just do a follow-up PR.

xylar:
@forsyth2, I don't know how the testing works but it won't be possible to actually run the `high_res_v1` campaign on Chrysalis because the necessary nodes (and for that matter the data) are only on Cori.  If the testing is just making sure the job script gets generated correctly and isn't actually trying to run analysis, that's probably fine on Chrysalis, though.

Also, I don't personally think the tutorial is the most intuitive place to have a note about the `high_res_v1` campaign.  Would it make sense (perhaps in another PR) to add more to the documentation about the `campaign` option in general?  It would seem important to document the contents of each `cfg` file in the same way `default.ini` is documented (these will be hard to find if folk are running from E3SM-Unified).  And for `high_res_v1`, we would want something about which simulations it is appropriate for (e.g. simulations with the `oRRS18to6` ocean and sea-ice mesh).

forsyth2:
> If the testing is just making sure the job script gets generated correctly and isn't actually trying to run analysis, that's probably fine on Chrysalis, though.

Yes, that's correct.

> Would it make sense (perhaps in another PR) to add more to the documentation about the campaign option in general?

That's a good idea. I can make a PR for that.

xylar:
Okay, thanks for answering my questions.  I think this is fine to merge.

PR #204, opened 2022-03-04
Update parameters

### LINKED ISSUE #194: Fix inheritance in configuration file
Specifying `mapping_file = /home/ac.zender/data/maps/map_ne30pg2_to_cmip6_180x360_aave.20200201.nc` under `[default]` results in creating `<output>/post/atm/native` whereas specifying the mapping file under both `[climo]` and `[ts]` results in creating `<output>/post/atm/180x360_aave/`.

It's likely the mapping file specified at https://github.com/E3SM-Project/zppy/blob/main/zppy/templates/default.ini#L20 and https://github.com/E3SM-Project/zppy/blob/main/zppy/templates/default.ini#L44 ends up overriding the value set in `[default]`.

Possible solution: move parameters as high up in the `default.ini` hierarchy as makes sense.

forsyth2:
Update parameters. Resolves #194.

`default.ini`:
- [x] Properly place parameters in the inheritance hierarchy of `default.ini`
- [x] Add comments from https://e3sm-project.github.io/zppy/_build/html/main/parameters.html directly above the parameters in `default.ini`
- [x] Alphabetize parameters in each section of `default.ini`

Tests/docs:
- [x] On https://e3sm-project.github.io/zppy/_build/html/main/parameters.html, direct users to `default.ini` rather than keeping a duplicate list of parameters in the docs
- [x] Alphabetize/update parameters for `test_bash_generation.py`
- [x] Alphabetize/update parameters for `test_camapign.py`
- [x] Alphabetize/update parameters for `test_complete_run.py`
- [x] Get unit tests passing
- [x] Check if `[task]` defaults in `default.ini` override `[default]` defaults as they would in a user’s `cfg`. Do this by creating `test_defaults.py`. Also document how to update the expected files.

After merging:
- [ ] Update expected files for tests and run tests again


forsyth2:
@golaz Just FYI, no action needed. This is a nice organizational improvement to our parameters. 

Unfortunately, there are 5 cases where I had to decide between 1) specifying a default in multiple sections or 2) specifying a default high up in the inheritance hierarchy and accepting that one section will *always* override the value. I went with option (2) because even if we don't set the default in the highest-up section, a user could still do so in their configuration file, so we might as well do the same.

Those cases are:
- `[climo]` sets `nodes = integer(default=4)`, overriding `nodes = integer(default=1)`
- `[climo]` sets `vars = string(default="")`, overriding `vars = string(default="FSNTOA,FLUT,FSNT,FLNT,FSNS,FLNS,SHFLX,QFLX,TAUX,TAUY,PRECC,PRECL,PRECSC,PRECSL,TS,TREFHT,CLDTOT,CLDHGH,CLDMED,CLDLOW,U")`
- `[tc_analysis]` sets `input_files = string(default="eam.h2")`, overriding `input_files = string(default="eam.h0")`
- `[mpas_analysis]` sets `walltime = string(default="06:00:00")`, overriding `walltime = string(default="02:00:00")`
- `[global_time_series]` sets `input_subdir = string(default="archive/ocn/hist")`, overriding `input_subdir = string(default="archive/atm/hist"`

PR #227, opened 2022-03-23
Bundle jobs

### LINKED ISSUE #189: Add option to bundle zppy jobs
This is a follow-up to https://github.com/E3SM-Project/zppy/issues/123. At the time, we did not know whether this would be a hypothetical or real issue, so no action was taken.

However, the current queue policy of NERSC makes it very difficult to execute the large number of batch jobs generated by zppy. A new restriction was put in place that causes **only two jobs to age at once**. zppy might be submitting dozens of jobs, but only the first two will age. It's only after they start running, that two other jobs will age. This makes the NERSC queue time even worse.

As an alternate option, we should introduce a "bundle" option for running zppy jobs. This change would introduce one new keyword in the configuration file: `bundle` which would be a string.

- `bundle` can be defined at any level: default, section or subsection.
- Default value is `""` which refers to current behavior. Backward compatibility is maintained.
- Setting `bundle` to a non empty string will cause all tasks impacted to be bundled together and submitted as a single script.
- User has great control over bundling granularity by defining `bundle` at different levels, and creating an arbitrary number of bundle jobs (identified by different strings).

Implementation details to follow.

forsyth2:
Bundle jobs. Resolves #189.

- [ ] Code review
- [ ] Address comments
- [ ] Test on NERSC (after #233 is merged)
- [ ] Merge
- [ ] Update expected files for integration tests

### DIFF
--- a/docs/source/dev_guide/testing.rst
+++ b/docs/source/dev_guide/testing.rst
@@ -20,6 +20,7 @@ Integration tests must be run on an LCRC machine. Run all integration tests by d
     .. code::
 
         pip install . # Install your changes (`python -m pip install .` also works)
+	# Run steps in the section below ("Commands to run before running integration tests")
         python -m unittest tests/integration/test_*.py # Run all integration tests
 
 Commands to run before running integration tests
@@ -36,6 +37,19 @@ Before running ``tests/integration/test_complete_run.py`` run the following:
        rm -rf /lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2/zppy_test_complete_run_www/v2.LR.historical_0201
        rm -rf /lcrc/group/e3sm/ac.forsyth2/zppy_test_complete_run_output/v2.LR.historical_0201/post
        zppy -c tests/integration/test_complete_run.cfg
+
+Before running ``tests/integration/test_bundles.py`` run the following:
+
+    .. code::
+
+       rm -rf /lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2/zppy_test_bundles_www/v2.LR.historical_0201
+       rm -rf /lcrc/group/e3sm/ac.forsyth2/zppy_test_bundles_output/v2.LR.historical_0201/post
+       zppy -c tests/integration/test_bundles.cfg
+       # bundle1 and bundle2 should run. After they finish, run:
+       rm /lcrc/group/e3sm/ac.forsyth2/zppy_test_bundles_output/v2.LR.historical_0201/post/scripts/bundle3.bash
+       # (If this file isn't deleted, zppy will fail because it assumes bundle3 is already running).
+       zppy -c tests/integration/test_bundles.cfg
+       # bundle3 and ilamb_run should run
 
 Commands to run to replace outdated expected files
 --------------------------------------------------
@@ -85,7 +99,24 @@ To replace the expected images for ``test_complete_run.py`` run the following:
        cd <top level of zppy repo>
        # Rerun test
        python -m unittest tests/integration/test_complete_run.py
-       
+
+To replace the expected images for ``test_bundles.py`` run the following:
+
+    .. code::
+      
+       # rm -rf /lcrc/group/e3sm/public_html/zppy_test_resources/expected_bundles
+       # Your output will now become the new expectation.
+       # Copy output so you don't have to rerun zppy to generate the output.
+       # cp -r /lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2/zppy_test_bundles_www/v2.LR.historical_0201 /lcrc/group/e3sm/public_html/zppy_test_resources/expected_bundles
+       # mkdir -p /lcrc/group/e3sm/public_html/zppy_test_resources/expected_bundles/bundle_files
+       # cp -r /lcrc/group/e3sm/ac.forsyth2/zppy_test_bundles_output/v2.LR.historical_0201/post/scripts/bundle*.bash /lcrc/group/e3sm/public_html/zppy_test_resources/expected_bundles/bundle_files
+       # cd /lcrc/group/e3sm/public_html/zppy_test_resources/expected_bundles
+       # This file will list all the expected images.
+       # find . -type f -name '*.png' > ../image_list_expected_bundles.txt
+       cd <top level of zppy repo>
+       # Rerun test
+       # python -m unittest tests/integration/test_bundles.py
+
 Automated tests
 ===============
 
--- a/zppy/__main__.py
+++ b/zppy/__main__.py
@@ -1,6 +1,7 @@
 import argparse
 import errno
 import os
+from typing import List
 
 from configobj import ConfigObj
 from validate import Validator
@@ -13,9 +14,11 @@ from zppy.ilamb_run import ilamb_run
 from zppy.mpas_analysis import mpas_analysis
 from zppy.tc_analysis import tc_analysis
 from zppy.ts import ts
+from zppy.utils import Bundle, submitScript
 
 
-def main():
+# FIXME: C901 'main' is too complex (19)
+def main():  # noqa: C901
 
     # Command line parser
     parser = argparse.ArgumentParser(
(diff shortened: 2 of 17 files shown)


zppy/ilamb_run.py
```diff
@@ -19,7 +19,7 @@ def ilamb_run(config, scriptDir):
     # --- List of ilamb_run tasks ---
     tasks = getTasks(config, "ilamb_run")
     if len(tasks) == 0:
-        return
+        return []
```
forsyth2, line 22: Note: should be `return existing_bundles`

zppy/templates/default.ini
```diff
@@ -23,6 +25,7 @@ input = string
 input_files = string(default="eam.h0")
 # The specific subdirectory with the atmospheric data
 input_subdir = string(default="archive/atm/hist")
+# machine -- NOTE: this parameter is created internally
```
forsyth2, line 28: Just adding this comment for `machine`, `scriptDir`, and `templateDir` -- these parameters existed before this pull request. `bundle` is the only new parameter.

zppy/utils.py
```diff
@@ -140,9 +143,135 @@ def getComponent(input_files):
 
 
 # -----------------------------------------------------------------------------
-def submitScript(scriptFile, dependFiles=[], export="ALL"):
+# Bundles
+class Bundle(object):
```
forsyth2, line 147: I ended up going with a class over a dictionary after all. There was a lot of information to process.

zppy/utils.py
```diff
@@ -140,9 +143,135 @@ def getComponent(input_files):
...
+        self.bundle_name: str = c["bundle"]
+        self.dry_run: bool = c[
+            "dry_run"
+        ]  # Value is taken from the first task in the bundle
+        self.script_dir: str = c[
+            "scriptDir"
+        ]  # Value is taken from the first task in the bundle
+
+        self.bundle_file: str = f"{self.script_dir}/{self.bundle_name}.bash"
+        self.bundle_output: str = self.bundle_file.strip(".bash") + ".o%j"
```
forsyth2, line 158: This literally ends with `".o%j"`, not the `o<job-id>` extension.

zppy/utils.py
```diff
@@ -140,9 +143,135 @@ def getComponent(input_files):
...
+
+        self.dependencies_in_bundle_file: Set[str] = set()
+        self.dependencies_not_in_bundle_file: Set[str] = set()
+
+        self.export: str = "NONE"
+
+    def create_header(self, scriptFile):
+        if os.path.exists(self.bundle_file):
+            error_message = f"create_header is being applied to an existing bundle: {self.bundle_name}. If {self.bundle_name} was running and you re-ran zppy, that is the likely cause of this error. If {self.bundle_name} is not currently running, delete {self.bundle_file} and run zppy again."
+            # By failing, zppy is prevented from restarting / overwriting work the bundle is currently doing.
```
forsyth2, line 168: This is the error I mentioned in `docs/source/dev_guide/testing.rst`.

zppy/utils.py
```diff
@@ -140,9 +143,135 @@ def getComponent(input_files):
...
+        # Create header or structure to keep info for the header.
+        with open(self.bundle_file, "w") as destination_file:
+            destination_file.write("#!/bin/bash\n")
+            with open(scriptFile, "r") as source_file:
+                for line in source_file:
+                    # Copy to bundle_file all lines from script_file that contain "SBATCH"
+                    if re.search("SBATCH", line):
+                        if re.search("--output", line):
+                            destination_file.write(
+                                f"#SBATCH --output={self.bundle_output}\n"
```
forsyth2, line 180: Output goes to a single output file for the bundle, not one for each task in the bundle. However, each task in the bundle keeps its own status file.

zppy/utils.py
```diff
@@ -140,9 +143,135 @@ def getComponent(input_files):
...
+                            )
+                        elif re.search("--job-name", line):
+                            destination_file.write(
+                                f"#SBATCH --job-name={self.bundle_name}\n"
```
forsyth2, line 184: We want to be sure to name the job with the bundle's name, not the first task's name.

zppy/utils.py
```diff
@@ -140,9 +143,135 @@ def getComponent(input_files):
...
+                            )
+                        else:
+                            destination_file.write(line)
+                # If any script fails, no new ones should try to run.
+                # It's possible the failed script is a dependency for later scripts.
```
forsyth2, line 189: Set the script to fail if any task fails, since there may be dependencies.

zppy/utils.py
```diff
@@ -140,9 +143,135 @@ def getComponent(input_files):
...
+            print("  None")
+
+
+def handle_bundles(c, scriptFile, export, dependFiles=[], existing_bundles=[]):
+    bundle_name = c["bundle"]
+    if bundle_name == "":
+        return existing_bundles
+    for b in existing_bundles:
+        if b.bundle_name == bundle_name:
+            # This bundle already exists
```
forsyth2, line 254: This prevents us from creating a bundle that already exists.

zppy/utils.py
```diff
@@ -140,9 +143,135 @@ def getComponent(input_files):
...
+        # If we're in this block, then we did not `break` from the for-loop.
+        # So, the bundle does not already exist
+        bundle = Bundle(c)
+        bundle.create_header(scriptFile)
+        existing_bundles.append(bundle)
+    bundle.handle_dependencies(dependFiles)
+    bundle.add_script(scriptFile)
+    if export == "ALL":
+        # If one task requires export="ALL", then the bundle script will need it as well
+        bundle.export = export
```
forsyth2, line 267: Since we're only launching one job rather than many, we have to make a decision on the value for `export`. I decided that if *any* task required `export="ALL"`, then the whole bundle would.

zppy/utils.py
```diff
@@ -140,9 +143,135 @@ def getComponent(input_files):
...
+    return existing_bundles
+
+
+# -----------------------------------------------------------------------------
+def submitScript(scriptFile, export, dependFiles=[]):
```
forsyth2, line 272: `export` is now a required parameter

zppy/__main__.py
```diff
@@ -97,29 +100,38 @@ def main():
...
 
     # ilamb_run tasks
-    ilamb_run(config, scriptDir)
+    existing_bundles = ilamb_run(config, scriptDir, existing_bundles)
+
+    for b in existing_bundles:
+        b.display_dependencies()
+        if not b.dry_run:
+            submitScript(
+                b.bundle_file, b.export, dependFiles=b.dependencies_not_in_bundle_file
```
forsyth2, line 133: Regarding my comment in `docs/source/dev_guide/testing.rst`: when we submit `bundle3`, it will have a dependency listed in `dependencies_not_in_bundle_file` which is run as part of `bundle1`. As the code currently is, however, it has no idea that this dependency is in `bundle1`.

forsyth2 (commented):
@golaz This is ready for review. I've included comments explaining the code. My suggested order of review is:

1. `docs/` and `tests/` (to get context on how `zppy` is run with bundles)
2. `zppy/__main__.py` and `zppy/utils.py` (for the majority of the code that enables the bundle functionality)
3. Other files in `zppy`: the task files & `zppy/templates/default.ini`

docs/source/dev_guide/testing.rst
```diff
@@ -37,6 +38,19 @@ Before running ``tests/integration/test_complete_run.py`` run the following:
...
 
+Before running ``tests/integration/test_bundles.py`` run the following:
+
+    .. code::
+
+       rm -rf /lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2/zppy_test_bundles_www/v2.LR.historical_0201
+       rm -rf /lcrc/group/e3sm/ac.forsyth2/zppy_test_bundles_output/v2.LR.historical_0201/post
+       zppy -c tests/integration/test_bundles.cfg
+       # bundle1 and bundle2 should run. After they finish, run:
+       rm /lcrc/group/e3sm/ac.forsyth2/zppy_test_bundles_output/v2.LR.historical_0201/post/scripts/bundle3.bash
```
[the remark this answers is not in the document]
forsyth2: @golaz a few options included in the new second commit:

Simply remove the `FileExistsError`: tests still pass except for the bash file content test. Basically, if you rerun `zppy` while `bundle1` is running, `bundle1.bash` gets overwritten (so it doesn't have the lines to run scripts that already have status files with "OK"). Presumably work begins twice on scripts that have not yet been run.

**Check for existence of a status file (e.g., `bundle1.status`) rather than a bash file: this seems to be the best solution.** 
- Downside is if `bundle1` is still running but has finished the task that `bundle3` needs, `zppy` will error out before having a chance to run `bundle3` (it will try to re-run `bundle1`, which is currently running, before getting to `bundle3`).
- Another downside is that it takes a few seconds for the status files to get created, so it's still possible to launch two identical jobs if you re-run `zppy` fast enough.
- Upside is that there is no need to delete `bundle3.bash` or any other files anymore -- `bundle3.status` doesn't get created until after it actually starts running.
- Tests still pass with this change, aside from the bash file content test (will need to include the line printing to the bundle's status file).

Lastly, I included, in a commented out section, code to print an error status if a particular script fails: while this would allow us to have a status simply beyond "the bundle has started", it does add of a lot lines to the scripts, for not much added value, I think.

golaz (changes requested):
@forsyth2,

Very nice work on this PR. I've made a few more modifications to your PR. The changes are available on a separate branch: https://github.com/E3SM-Project/zppy/tree/bundle-jobs. Maybe you can import them into this PR.

My changes add the following:

- Bundle bash scripts are now constructed from a Jinja2 template. This will facilitate future changes.
- Ability to explicitly declare bundle jobs in the cfg file via new section and sub-sections.

```
[bundle]

  [[ bundle2 ]]
  nodes=2
```
  
- Some code restructuring.

Notes:

- black doesn't work. You branched before the changes to fix black were merged in. I used `--no-verify`
- The integration tests run fine, but I have not updated the expected output, so one reports a failure.
- For the complete integration tests, the status files for mpas_analysis and tc have the weird 'NING <jobID>' status.

forsyth2:
Closing in favor of #243.

Issue #165, opened 2021-11-04
Optimize `ncclimo` calls


forsyth2:
Optimize `ncclimo` calls.

`zppy/templates/climo.bash` Monthly climatology:
- `--mem_mb=0` is only necessary on Cori, but is ok to use elsewhere. This can be removed after the release of E3SM Unified 1.5.1 (which will include `zppy v1.1.0`). See @czender's comment on #136.
- `--thr=1` is not necessary for low-res. In fact, it hurts low-res performance. @chengzhuzhang and @xylar may consider a hi-res configuration. Related to #9?
- Replace `--model` (`-m`) with `--prc_typ` (`-P`).

`zppy/templates/climo.bash` Diurnal cycle climatology:
- First two bullets above apply here as well.

`zppy/templates/ts.bash`:
- Use `--area` rather than `--area_nm` (same option, just a different name)
- Test with `dpf` and `tpd` removed. Those two options shouldn't be necessary.

Note: to see all `ncclimo` options, just run `ncclimo` with no options.

czender:
I need to add one further comment about this: Replace --model (-m) with --prc_typ (-P). This always works fine for v2 simulations. However, if model is specified as "eam"  (e.g., --prc_typ=eam) then ncclimo will expect the raw filenames to contain "eam". This only became true late in v2 development. If the raw filenames contain "cam" then the model should be specified as "cam", e.g., --prc_typ=cam.

chengzhuzhang:
Mostly resolved in #228 
Can revisit to further optimize later.

PR #197, opened 2022-02-17
ILAMB task

chengzhuzhang:
This PR is to add capability to run ILAMB package for land model bench-marking. Close #133, #134. 

Example configuration file:
```
[default]
input = /lcrc/group/e3sm/ac.forsyth2/E3SM_simulations/20210122.v2_test01.piControl.ne30pg2_EC30to60E2r2-1900_ICG.chrysalis
input_subdir = archive/atm/hist
output = /lcrc/group/e3sm/ac.zhang40/zppy_tests/zppy_ilamb_324_tr1 
case = 20210122.v2_test01.piControl.ne30pg2_EC30to60E2r2-1900_ICG.chrysalis
www = /lcrc/group/e3sm/public_html/diagnostic_output/ac.zhang40/zppy_ilamb
partition = compute
# TODO: will add a configuration template for "BGC" campaign
campaign = "water_cycle"
environment_commands = "source /lcrc/soft/climate/e3sm-unified/load_latest_e3sm_unified_chrysalis.sh"

# Regridded monthly time series files in 10 year chunks
[ts]
active = True
years = "1:10:10",
ts_fmt = "cmip"

  # Monthly land time series
  [[ land_monthly ]]
  input_subdir = "archive/lnd/hist"
  input_files = "elm.h0"
  frequency = "monthly"
  mapping_file = /home/ac.zender/data/maps/map_ne30pg2_to_cmip6_180x360_aave.20200201.nc
  # TODO: add more variables requested by BGC
  vars = "LAISHA,LAISUN"

  [[ atm_monthly_180x360_aave ]]
  input_subdir = "archive/atm/hist"
  input_files = "eam.h0"
  frequency = "monthly"
  mapping_file = /home/ac.zender/data/maps/map_ne30pg2_to_cmip6_180x360_aave.20200201.nc
  # TODO: add more variables requested by BGC
  vars = "PRECC,PRECL"

[ilamb_run]
active = True
years = "1:10:10",
ts_num_years = 10
short_name = '20210122.v2_test01'
  [[ land_monthly ]]
```

- [x] Integrate in e3sm_to_cmip function for additional parameter to generate CMIP-like time-series for land and atmosphere variables, that ilamb_run task required.
- [x] Integrate in ilamb_run task.
- [ ] TODO: Include equivalent mpi call for supported machines.
- [ ] TODO: Copy files to web servers.
- [x] TODO: figure out path for ILAMB obs data.

External development for e3sm_to_cmip
[...]

PR #228, opened 2022-03-23
revise ncclimo calls

### LINKED ISSUE #220: ncclimo call may need to be fixed for land variable regridding
Better discription of this issue:
> The figure looks about right in that it has color in all the right places where `landfrac > 0.0`. So the "landmask" appears to be correct. However, the scripts were never told that this was ELM data so every native gridcell was assumed to have `landfrac == 1.0` and the SGS routines were not employed during regridding. The differences are typically not large (a few percent at most) in the global mean, and so difficult to detect by eye.
> 

An example provided on https://github.com/E3SM-Project/zppy/issues/220#issuecomment-1076604080
image

/lcrc/group/e3sm/ac.forsyth2/E3SMv2/v2.LR.historical_0201/post/lnd/180x360_aave/ts/monthly/5yr/FSH_185001_185412.nc
Screen Shot 2022-03-18 at 9 58 50 AM

chengzhuzhang:
This PR closes #220 #165 
@czender Would you please help review this PR. 
In the metadata of newly generated land variable data, there is a new line in `history`: ncap2 -O -s area*=6.37122e6^2/1.0e6;area@long_name=\"Gridcell area\";area@units=\"km^2\" output/LAISUN_000101_002512.nc output/LAISUN_000101_002512.nc

Does this mean the sgs method is applied successful?

A separate question for e3sm2cmip. It's tricky to identify if sgs_frac is applied in cmorized time series. The early versions of land data has following in the global attribute: "
```
:ncclimo_generation_command = "ncclimo --var=${var} -7 --dfl_lvl=1 --no_cll_msr --no_frm_trm --no_stg_grd --yr_srt=1 --yr_end=500 --ypf=25 --map=map_ne30np4_to_cmip6_180x360_aave.20181001.nc " ;
:ncclimo_version = "4.8.1-alpha04""
```
But for newly cmorized data, there is no such history retained...and perhaps we can use one of the temperature variable (i.g. soil temperature) and it's temperature range as an indicator if sgs is applied?

### DIFF
--- a/zppy/templates/climo.bash
+++ b/zppy/templates/climo.bash
@@ -27,8 +27,6 @@ cd ${workdir}
 ncclimo \
 --case={{ case }} \
 --jobs=${SLURM_NNODES} \
---mem_mb=0 \
---thr=1 \
 {%- if exclude %}
 -n '-x' \
 {%- endif %}
@@ -46,7 +44,7 @@ ncclimo \
 --output=trash \
 --regrid=output \
 {%- endif %}
---model={{ input_files.split(".")[0] }}
+--prc_typ={{ input_files.split(".")[0] }}
 
 {% elif frequency.startswith('diurnal') %}
 
@@ -89,8 +87,6 @@ ls {{ case }}.{{ input_files }}.????-*.nc > input.txt
 cat input.txt | ncclimo \
 --case={{ case }}.{{ input_files }} \
 --jobs=${SLURM_NNODES} \
---mem_mb=0 \
---thr=1 \
 {%- if exclude %}
 -n '-x' \
 {%- endif %}
--- a/zppy/templates/climo.bash
+++ b/zppy/templates/climo.bash
@@ -27,6 +27,7 @@ cd ${workdir}
 ncclimo \
 --case={{ case }} \
 --jobs=${SLURM_NNODES} \
+--thr=1 \
 {%- if exclude %}
 -n '-x' \
 {%- endif %}
@@ -44,7 +45,11 @@ ncclimo \
 --output=trash \
 --regrid=output \
 {%- endif %}
+{%- if input_files.split(".")[0] == 'cam' or input_files.split(".")[0] == 'eam'%}
 --prc_typ={{ input_files.split(".")[0] }}
+{%- else -%}
+--prc_typ=sgs
+{%- endif %}
 
 {% elif frequency.startswith('diurnal') %}
 
@@ -87,6 +92,7 @@ ls {{ case }}.{{ input_files }}.????-*.nc > input.txt
 cat input.txt | ncclimo \
 --case={{ case }}.{{ input_files }} \
 --jobs=${SLURM_NNODES} \
+--thr=1 \
 {%- if exclude %}
 -n '-x' \
 {%- endif %}
--- a/zppy/templates/ts.bash
+++ b/zppy/templates/ts.bash
@@ -73,7 +73,7 @@ cat input.txt | ncclimo \
 {%- elif mapping_file == 'glb' -%}
 -o output \
 --glb_avg \
---area_nm={{ area_nm }} \
+--area={{ area_nm }} \
 {%- else -%}
 --map={{ mapping_file }} \
 -o trash \
@@ -84,7 +84,13 @@ cat input.txt | ncclimo \
 --dpf={{ dpf }} \
 --tpd={{ tpd }} \
 {%- endif %}
+{%- if input_files.split(".")[0] == 'cam' or input_files.split(".")[0] == 'eam'%}
 --prc_typ={{ input_files.split(".")[0] }}
+{%- else -%}
+--prc_typ=sgs
+{%- endif %}
+
+
 
 if [ $? != 0 ]; then
   cd {{ scriptDir }}
--- a/zppy/templates/ts.bash
+++ b/zppy/templates/ts.bash
@@ -84,6 +84,7 @@ cat input.txt | ncclimo \
 --dpf={{ dpf }} \
 --tpd={{ tpd }} \
 {%- endif %}
+--prc_typ={{ input_files.split(".")[0] }}
 
 if [ $? != 0 ]; then
   cd {{ scriptDir }}


czender (changes requested):
According to this change what are the possible arguments that will be passed to the `--prc_typ` option? Valid arguments are described here http://nco.sf.net/nco.html#prc_typ.

The --thr=1 option was added to accomodate HR climos on Cori 96 GB nodes, IIRC, because that is the least amount of RAM on any major system. Otherwise thr_nbr will default to 2. This PR should be tested with HR data on Cori before this change is made. 

The additional ncap2 command in the history attribute is because you changed to `-P elm`. As I tried to explain earlier, this causes the regridder to try to produce an output indistinguishable (except for the grid) from native ELM output. And this means rescaling the area field back to ELM-native km2 instead of steradians. Personally I prefer area in steradians, which is the CMIP standard unit for area. Use `'-P sgs' `or `'--sgs_frc=landfrac' instead of `-P elm` to retain the area in steradians. Please decide which form to use based on the stakeholders involved.

czender:
Regarding your second question, on `e3sm2cmip`, the quoted ncclimo command indicates that SGS regridding was not performed. With newly cmorized data where the command provenance (`history` attribute) has been stripped off/down, then yes, examination of a temperature variable is a perfectly good way to discern whether SGS algorithms were employed. In fact the temperature method is the better method than the `history` attribute because there are corner cases one can imagine where someone might have built a map-file that accounts for SGS regridding and renamed it to the same name as a mapfile that does not account for SGS regridding and then the `history` attribute would give the wrong impression. The proof is always in the temperature field.

chengzhuzhang:
Thank you!
> According to this change what are the possible arguments that will be passed to the `--prc_typ` option? Valid arguments are described here http://nco.sf.net/nco.html#prc_typ.
> 
Possible input arguments I have seen include, `elm`, `clm2`, `eam`, `cam`. 

> The --thr=1 option was added to accomodate HR climos on Cori 96 GB nodes, IIRC, because that is the least amount of RAM on any major system. Otherwise thr_nbr will default to 2. This PR should be tested with HR data on Cori before this change is made.
> 
I assumed that the high res configuration has a customized setting for all jobs, but it looks [mpas-a] has it but the [ts] [climo] settings were not included yet, I will keep the --thr=1 as a general setting for now.

> The additional ncap2 command in the history attribute is because you changed to `-P elm`. As I tried to explain earlier, this causes the regridder to try to produce an output indistinguishable (except for the grid) from native ELM output. And this means rescaling the area field back to ELM-native km2 instead of steradians. Personally I prefer area in steradians, which is the CMIP standard unit for area. Use `'-P sgs' `or `'--sgs_frc=landfrac' instead of `-P elm` to retain the area in steradians. Please decide which form to use based on the stakeholders involved.

I think it makes sense to use the CMIP standard unit for area to make our data streams consistent, in terms of implementation. In the new commit,
if input is neither `eam` or `cam`, i set  `-P sgs`. Please review again...

czender (approved):
Changes look good to me.

czender:
FYI `prc_typ` is coded to recognize model names not versions, so `clm` is recognized and `clm2` is not. The workaround for `clm2` is `-P elm --mdl_nm=clm2`, i.e., the model name as it appears in the filename string for history files is copied from the `prc_typ` argument unless `mdl_nm` is explicitly specified. Also `prc_typ=eam` or `cam` are recognized though they are currently no-ops, i.e., hooks for if/when we need to add some special processing for EAM in the future. The defaults for each processing type are, of course, open for debate and change.

chengzhuzhang:
Okay, got it, I noticed `clm` in the doc in the list of possible arguments of `prc_typ`. So the implemented logic to use `sgs` if not `eam or `cam` should take account of this variation.

chengzhuzhang:
Thank you for also testing!
My tests went okay. I think it is just a formatting issue, but good to fix.

chengzhuzhang:
I will fix this in the ilamb PR

chengzhuzhang:
This commit https://github.com/E3SM-Project/zppy/pull/197/commits/2425aa6560e0944346b98d01fe05b23c0495d981 should resolve the formatting problem.

czender:
FYI while adding the spot-check for SGS invocation of ELM files into the regridder, I discovered that a bug related to this PR had crept into the `ncremap` used in E3SMU 1.6.0. The effect of the bug is to essentially ignore the `-P elm` (and similar) options. I am still investigating the dimensions of the problem and will update this tomorrow. However, it seems likely that the minimal workaround will involve changing this PR. Worst case is we may need to issue an "emergency" update E3SMU 1.6.0.1 or somesuch. Fingers crossed that's not necessary.

chengzhuzhang:
Thank you for updating. I think this PR will most likely be used in operation with the next E3SMU major release. Hope a minimal workaround fix will work..

czender:
The current E3SMU (1.6.0) will produce incorrect results when SGS regridding is invoked by calling `ncclimo` (or `ncremap`) with the `-P prc_typ` option alone. I intended that option to be sufficient by itself, but a bug I introduced in `ncremap` prevents that option from being fully parsed. As a result, the `-P prc_typ` does not currently trigger SGS regridding. However, it is harmless to leave it in any EZ commands that use it, and it remains essential in processing MPAS data, and it should be retained in those commands because all this will work as intended as of the NCO (5.0.7, which I am now finalizing) and eventually the next E3SMU. 
The workaround to obtain proper SGS normalization with NCO 5.0.6 and E3SMU 1.6.0 is to explicitly invoke `ncclimo` (or `ncremap`) with the `--sgs_frc=landfrac` (for ELM/CLM) or `--sgs_frc=timeMonthly_avg_iceAreaCell` (for MPASSI) options in addition to any `-P prc_typ` options already employed. To be clear, this workaround only need be applied to processing of ELM (not EAM) and MPASSI output. @chengzhuzhang and @xylar would you please confirm receipt/understanding of this issue. I'm happy to answer any questions or help with any PRs necessary to implement this workaround. My apologies for finding a bug on a Friday in the first place!

czender:
Good question. To generate+regrid ELM timeseries with area in steradians, use `--sgs_frc=landfrac` with or without (it doesn't matter) `-P sgs`. To generate+regrid ELM climos with area in steradians, do the same as for timeseries and, in addition, add `--mdl_nm=elm` or `--mdl_nm=clm2` depending on the string used in the filenames.

PR #237, opened 2022-05-04
Model-vs-model improvements

### LINKED ISSUE #223: model-vs-model improvements
model-vs-model (#195) improvements:

- Have `test_complete_run.cfg` run on a different set of years for the reference -- then the difference plots will actually be meaningful.
- Make assumptions of what the other reference data paths should be given `reference_data_path`. (We can leave those parameters but they should be automatically filled in if possible).
- `gauges_path` could be stored in the E3SM Diags repo or be specified by Mache; then it would not necessarily have to be a parameter.

forsyth2:
Model-vs-model improvements. Resolves #223. Improvement to #195.

### DIFF
--- a/tests/integration/test_complete_run.cfg
+++ b/tests/integration/test_complete_run.cfg
@@ -66,7 +66,7 @@ years = "1850:1854:2",
 active = True
 scratch = "/lcrc/globalscratch/ac.forsyth2"
 walltime = "00:30:00"
-years = "1850:1852:2",
+years = "1850:1854:2",
 
 [e3sm_diags]
 active = True
@@ -104,22 +104,18 @@ years = "1850:1854:2", "1850:1854:4",
   climo_diurnal_subsection = "atm_monthly_diurnal_8xdaily_180x360_aave"
   climo_subsection = "atm_monthly_180x360_aave"
   diff_title = "Difference"
-  gauges_path = "/lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/obs/time-series/GSIM/GSIM_catchment_characteristics_all_1km2.csv"
   ref_final_yr = 1851
   ref_name = "v2.LR.historical_0201"
   ref_start_yr = 1850
   ref_years = "1850-1851",
   reference_data_path = "/lcrc/group/e3sm/ac.forsyth2/zppy_test_complete_run_output/v2.LR.historical_0201/post/atm/180x360_aave/clim"
-  reference_data_path_climo_diurnal = "/lcrc/group/e3sm/ac.forsyth2/zppy_test_complete_run_output/v2.LR.historical_0201/post/atm/180x360_aave/clim_diurnal_8xdaily"
-  reference_data_path_tc = "/lcrc/group/e3sm/ac.forsyth2/zppy_test_complete_run_output/v2.LR.historical_0201/post/atm/tc-analysis_1850_1851"
-  reference_data_path_ts = "/lcrc/group/e3sm/ac.forsyth2/zppy_test_complete_run_output/v2.LR.historical_0201/post/atm/180x360_aave/ts/monthly"
-  reference_data_path_ts_rof = "/lcrc/group/e3sm/ac.forsyth2/zppy_test_complete_run_output/v2.LR.historical_0201/post/rof/native/ts/monthly"
   run_type = "model_vs_model"
   short_ref_name = "v2.LR.historical_0201"
   swap_test_ref = False
   tag = "model_vs_model"
   ts_num_years_ref = 2
   ts_subsection = "atm_monthly_180x360_aave"
+  years = "1852-1853",
 
 [mpas_analysis]
 active = True
--- a/tests/integration/test_complete_run.py
+++ b/tests/integration/test_complete_run.py
@@ -9,13 +9,18 @@ from PIL import Image, ImageChops, ImageDraw
 # Copied from E3SM Diags
 def compare_images(
     test,
+    missing_images,
     mismatched_images,
     image_name,
     path_to_actual_png,
     path_to_expected_png,
 ):
     # https://stackoverflow.com/questions/35176639/compare-images-python-pil
-    actual_png = Image.open(path_to_actual_png).convert("RGB")
+    try:
+        actual_png = Image.open(path_to_actual_png).convert("RGB")
+    except FileNotFoundError:
+        missing_images.append(image_name)
+        return
     expected_png = Image.open(path_to_expected_png).convert("RGB")
     diff = ImageChops.difference(actual_png, expected_png)
 
@@ -80,6 +85,7 @@ class TestCompleteRun(unittest.TestCase):
             "/lcrc/group/e3sm/public_html/zppy_test_resources/expected_complete_run"
         )
 
+        missing_images: List[str] = []
         mismatched_images: List[str] = []
 
         counter = 0
@@ -94,12 +100,23 @@ class TestCompleteRun(unittest.TestCase):
 
                 compare_images(
                     self,
+                    missing_images,
                     mismatched_images,
                     image_name,
                     path_to_actual_png,
                     path_to_expected_png,
                 )
 
+        if missing_images:
+            print("Missing images:")
+            for i in missing_images:
+                print(i)
+        if mismatched_images:
+            print("Mismatched images:")
+            for i in mismatched_images:
+                print(i)
+
+        self.assertEqual(missing_images, [])
         self.assertEqual(mismatched_images, [])
 
 
--- a/zppy/e3sm_diags.py
+++ b/zppy/e3sm_diags.py
@@ -84,9 +84,24 @@ def e3sm_diags(config, scriptDir):  # noqa: C901
                         "reference_data_path_ts_rof"
                     ] = f"{reference_data_path}/rof/native/ts/monthly"
                 if c["gauges_path"] == "":
-                    c[
-                        "gauges_path"
-                    ] = "/lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/obs/time-series/GSIM/GSIM_catchment_characteristics_all_1km2.csv"  # Define with Mache?
+                    gauges_path_suffix = (
+                        "time-series/GSIM/GSIM_catchment_characteristics_all_1km2.csv"
+                    )
+                    if c["machine"] == "compy":
+                        gauges_path_prefix = (
+                            "/compyfs/e3sm_diags_data/obs_for_e3sm_diags/"
+                        )
+                    elif c["machine"] == "cori":
+                        gauges_path_prefix = (
+                            "/global/cfs/cdirs/e3sm/e3sm_diags/obs_for_e3sm_diags/"
+                        )
+                    elif c["machine"] in ["anvil", "chrysalis"]:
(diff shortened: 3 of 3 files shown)


tests/integration/test_complete_run.cfg
```diff
@@ -104,22 +104,18 @@ years = "1850:1854:2", "1850:1854:4",
...
-  reference_data_path_tc = "/lcrc/group/e3sm/ac.forsyth2/zppy_test_complete_run_output/v2.LR.historical_0201/post/atm/tc-analysis_1850_1851"
-  reference_data_path_ts = "/lcrc/group/e3sm/ac.forsyth2/zppy_test_complete_run_output/v2.LR.historical_0201/post/atm/180x360_aave/ts/monthly"
-  reference_data_path_ts_rof = "/lcrc/group/e3sm/ac.forsyth2/zppy_test_complete_run_output/v2.LR.historical_0201/post/rof/native/ts/monthly"
   run_type = "model_vs_model"
   short_ref_name = "v2.LR.historical_0201"
   swap_test_ref = False
   tag = "model_vs_model"
   ts_num_years_ref = 2
   ts_subsection = "atm_monthly_180x360_aave"
+  years = "1852-1853",
```
forsyth2, line 118: Now comparing 1852-1853 and 1850-1851, rather than comparing the same year sets. So, now there is actually a diff to plot.

tests/integration/test_complete_run.py
```diff
@@ -9,13 +9,18 @@
...
     image_name,
     path_to_actual_png,
     path_to_expected_png,
 ):
     # https://stackoverflow.com/questions/35176639/compare-images-python-pil
-    actual_png = Image.open(path_to_actual_png).convert("RGB")
+    try:
+        actual_png = Image.open(path_to_actual_png).convert("RGB")
+    except FileNotFoundError:
+        missing_images.append(image_name)
```
forsyth2, line 22: I added this because the test was erroring out because I changed the year-set and hadn't updated the expected files yet.

zppy/e3sm_diags.py
```diff
@@ -64,6 +64,29 @@ def e3sm_diags(config, scriptDir):  # noqa: C901
                     c["ref_year1"],
                     c["ref_year2"],
                 )
+                reference_data_path = (
```
forsyth2, line 67: Now, these parameters are defined with `reference_data_path`. They can still be specified explicitly though.

zppy/e3sm_diags.py
```diff
@@ -64,6 +64,29 @@ def e3sm_diags(config, scriptDir):  # noqa: C901
...
+                        "reference_data_path_ts"
+                    ] = f"{reference_data_path}/atm/{c['grid']}/ts/monthly"
+                if c["reference_data_path_ts_rof"] == "":
+                    c[
+                        "reference_data_path_ts_rof"
+                    ] = f"{reference_data_path}/rof/native/ts/monthly"
+                if c["gauges_path"] == "":
+                    c[
+                        "gauges_path"
+                    ] = "/lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/obs/time-series/GSIM/GSIM_catchment_characteristics_all_1km2.csv"  # Define with Mache?
```
forsyth2, line 89: Should we leave this hard-coded for now? I'm not sure what would be involved to use Mache here. Wouldn't Mache have to define a `gauges_path` for each machine?

I think we'd also need to define Mache as a dependency in the `conda` environment files too.

forsyth2 (commented):
@chengzhuzhang @golaz I made the improvements suggested in our 3/18 meeting. Let me know if I should do anything else to improve upon #195 before the Unified release.

zppy/e3sm_diags.py
chengzhuzhang: hmmm, It is a little confusing to have to point to a obs path for `model vs model` run. I think the cleaner way is to include the gauge locations as a separate file, which should install with e3sm_diags. Before having the separation implement, I thank Mache can be used here to determine the obs path per machines.

chengzhuzhang (changes requested):
It seems this implementation would only work for chrysalis with obs path for streamflow hard-coded. Should change how guage locations retrieved from e3sm_diags to avoid have to refer to obs path, or use Mache to get machine specific obs paths.

zppy/e3sm_diags.py
golaz: +1 to use Mache for all the necessary paths in E3SM Diags

xylar: @forsyth2, the easiest might be for us to have a short chat.

Just add `mache >=1.3.2` as a requirement for `zppy` and you'll have it.  It's on conda-forge: https://github.com/conda-forge/mache-feedstock

There's no need to do a development environment for `mache` unless you need to add new functionality.  So far, I don't see any reason you would.

Then, you would get a `MachineInfo` object like described in the readme: https://github.com/E3SM-Project/mache/blob/main/README.rst.  On compute nodes, you need some way to know what machine you're on.  But that can be automatic for `zppy` because it always runs on login nodes, which `mache` will detect automatically.

I believe that E3SM Diags is now storing this data in the "sanctioned" place on each machine, within the shared diagnostics directory.  @chengzhuzhang, please correct me if that's not yet the case.  If so, you can get the diagnostics directory on a given machine most easily with:
``` python
from mache import MachineInfo

...
machine_info = MachineInfo()
diags_base = machine_info.config.get('diagnostics', 'base_path')
```
You can see what other config options are available by looking at an example machine:
https://github.com/E3SM-Project/mache/blob/main/mache/machines/chrysalis.cfg

If E3SM Diags is storing these files in its own place outside of `diags_base`, we're going to have a harder time and we need to discuss this.  But I could (reluctantly) add something new to `mache` specifically for E3SM Diags if that's the best way forward.

xylar: Just to clarify, the "sanctioned" location for diagnostic data is:
```
/lcrc/group/e3sm/public_html/diagnostics/
```
if it can be available to the public or:
```
/lcrc/group/e3sm/diagnostics_private/
```
if it's supposed to be available only to E3SM team members.  Then, the `mache` sync functionality (https://github.com/E3SM-Project/mache#syncing-diagnostics) that I just added can be used to (manually) sync this data to `diags_base` on each machine.

chengzhuzhang: Yes, the same streamflow data is here: /lcrc/group/e3sm/public_html/diagnostics/observations/Atm/time-series/GSIM/GSIM_catchment_characteristics_all_1km2.csv

I haven't used mache to populate the same /lcrc/group/e3sm/public_html/diagnostics/ directory on other machines. This is a good opportunity to unify the location . [so we don't need to maintain separate e3sm_diags data in another path on different machines.]

xylar: @chengzhuzhang, once LCRC is back up, I can run the sync script on all machines (except Acme1, as usual).

chengzhuzhang: Thank you @xylar ! i will take care of acme1 :)

xylar: @chengzhuzhang, I just synced on Anvil/Chrysalis, Andes, Compy, Cooley and Cori.  In all cases, there was nothing to sync.  I think I had told you not to worry about Acme1 until the next `mache` release and I think that advice still holds for now unless you want to test this PR there.

forsyth2:
@chengzhuzhang I added a second commit to handle the `gauges_path` better -- see https://github.com/E3SM-Project/zppy/pull/237/commits/6bd0d9a7d633a9aa56ef79c62072cd086677b4f6. If that looks good, I think this is good to merge.

@xylar Thanks for the instructions on using Mache. I was able to generate the following. @chengzhuzhang  -- would we want to have more paths added to the `Diagnostics` section (e.g., `obs_path`, `gauges_path`)? If so, we can schedule a time for the three of us to discuss the implementation.

```
Machine: chrysalis
  E3SM Supported Machine: True
  Compilers: intel, gnu
  MPI libraries: openmpi, impi
  OS: LINUX

E3SM-Unified:
  E3SM-Unified is not currently loaded
  Base path: /lcrc/soft/climate/e3sm-unified

Diagnostics:
  Base path: /lcrc/group/e3sm/diagnostics

Config options:
  [e3sm_unified]
    group = cels
    compiler = gnu
    mpi = openmpi
    base_path = /lcrc/soft/climate/e3sm-unified

  [diagnostics]
    base_path = /lcrc/group/e3sm/diagnostics

  [web_portal]
    base_path = /lcrc/group/e3sm/public_html/diagnostic_output
    base_url = https://web.lcrc.anl.gov/public/e3sm/diagnostic_output
    username = ac.forsyth2

  [parallel]
    system = slurm
    parallel_executable = srun
    cores_per_node = 128
    partitions = debug, compute, high

  [spack]
    modules_before = False
    modules_after = True
```

chengzhuzhang (approved):
i think it is fine as a temporal workaround for now for this PR. I would try to remove dependency of gauge_path for gauge location information in e3sm_diags. And better to use more Mache for future development.

forsyth2:
> I would try to remove dependency of gauge_path for gauge location information in e3sm_diags.

@chengzhuzhang Thanks for creating https://github.com/E3SM-Project/e3sm_diags/issues/588

> There's no need to do a development environment for mache unless you need to add new functionality. So far, I don't see any reason you would.

@xylar Even if I personally aren't adding new functionality, isn't the development environment necessary to get the latest commits from Mache (i.e., those that haven't been released yet)?

forsyth2:
>  I don't think there is new functionality you need, though.

@xylar In the future, that may prove useful if more paths are added for the diagnostics section for example. For this pull request however, it's not required.

xylar:
> @xylar In the future, that may prove useful if more paths are added for the diagnostics section for example. For this pull request however, it's not required.

There are certainly circumstances where that might be useful.  But I don't think we plan to add any new directories to the diagnostics section.  The idea is that *everything* we sync related to diagnostics is supposed to be in that directory.

forsyth2:
Ah ok, I understand, thanks for the clarification. In that case, @chengzhuzhang, we should coordinate on getting all the diagnostics info into that directory. Then we could actually replace
```
                    if c["machine"] == "compy":
                        gauges_path_prefix = (
                            "/compyfs/e3sm_diags_data/obs_for_e3sm_diags/"
                        )
                    elif c["machine"] == "cori":
                        gauges_path_prefix = (
                            "/global/cfs/cdirs/e3sm/e3sm_diags/obs_for_e3sm_diags/"
                        )
                    elif c["machine"] in ["anvil", "chrysalis"]:
                        gauges_path_prefix = "/lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/obs/"
```
with 
```
machine_info = MachineInfo()
# On chrysalis, this will be `/lcrc/group/e3sm/diagnostics`:
diags_base = machine_info.config.get('diagnostics', 'base_path')
gauges_path = os.path.join(diags_base,  "obs_for_e3sm_diags/time-series/GSIM/GSIM_catchment_characteristics_all_1km2.csv")
```

chengzhuzhang:
Yes, it has been done. We should rely on the single diagnostics path pointed by Mache. I will plan to clean up other e3sm_diags path on each machine and make symlinks to the default path.

forsyth2:
Great! Yes, I will need to update quite a few paths then. I just created #240 for this.

Issue #242, opened 2022-05-13
Corrupted status files on chrysalis (NING bug)


golaz:
This is likely a new issue that appeared after the monthly maintenance on Chrysalis this past Monday.

For certain zppy tasks (mpas_analysis, tc_analysis), the status files upon successful completion of a task are corrupt. The content of the status file should simply be
```
OK
```

but instead, it looks something like
```
OK
NING 176138
```
The bash line of code that updates the file looks like

```
echo 'OK' > /lcrc/group/e3sm/ac.golaz/zppy_test_complete_run_output/v2.LR.historical_0201/post/scripts/mpas_analysis_ts_1850-1860_climo_1855-1860.status
```
I don't understand how the status file can become corrupt and I have not been able to reproduce in a simple test (but the full task will repeatedly and consistently produce the erroneous status).

A simple workaround that appears effective is to first delete the file:

```
rm -f /lcrc/group/e3sm/ac.golaz/zppy_test_complete_run_output/v2.LR.historical_0201/post/scripts/mpas_analysis_ts_1850-1860_climo_1855-1860.status
echo 'OK' > /lcrc/group/e3sm/ac.golaz/zppy_test_complete_run_output/v2.LR.historical_0201/post/scripts/mpas_analysis_ts_1850-1860_climo_1855-1860.status
```

golaz:
Also tagging @xylar out of pure curiosity. Have you ever seen anything like that?

forsyth2:
This is a duplicate of #241. As this issue provides more info, I'm keeping this open and closing #241.

golaz:
Also tagging @rljacob. Do you know what was done during the Chrysalis maintenance last Monday? Any upgrades to bash or the file system? In this particular case, the bug is mostly harmless. But if it is a symptom of a larger issue, it could be quite serious (file integrity).

xylar:
I would say this is almost certainly a race condition.  This call:
https://github.com/E3SM-Project/zppy/blob/335b33ae55f1651b1a77aeeca937ad50aa9f7982/zppy/templates/mpas_analysis.bash#L23
most likely happens at the same time as this call:
https://github.com/E3SM-Project/zppy/blob/335b33ae55f1651b1a77aeeca937ad50aa9f7982/zppy/templates/mpas_analysis.bash#L388
But I can't immediately see why that would happen.

xylar:
I don't have much experience with bash redirects like this.  I use python almost exclusively for logging.

[...]

PR #243, opened 2022-05-15
Bundle jobs revision

### LINKED ISSUE #189: Add option to bundle zppy jobs

forsyth2:
Bundle jobs revision. Replaces #227. Resolves #189.


zppy/bundle.py
```diff
@@ -82,22 +80,23 @@ def add_task(self, scriptFile, dependFiles):
     def display_dependencies(self):
         print(f"Displaying dependencies for {self.bundle_name}")
         print("dependencies_internal:")
-        # Note to Ryan: I don't think te for else structure is working as expected
```
forsyth2: Possible conditions:
1. List in `for` block is empty.
2. `for` block completes without hitting `break`.
3. `for` block ends early because of `break`.

The `else` block runs in (1) and (2) but not in (3). It looks like I got confused here and thought the `else` block only ran in (1). That's why it worked in the `handle_bundles` `for-else` block; there's a `break`.

forsyth2 (commented):
@golaz This is ready for review.

The first commit is my initial work (combined 2 commits) from #227. The second commit is the combined changes of your 7 commits on https://github.com/E3SM-Project/zppy/tree/bundle-jobs. The third commit is my updates to your updates.

From your comments on #227:

> black doesn't work. You branched before the changes to fix black were merged in. I used --no-verify

I did branch before #236, but I rebased to include that commit, so I'm not sure why it wasn't working for you. In any case, I was able to run `pre-commit run --all-files`. The third commit includes the formatting updates.

> The integration tests run fine, but I have not updated the expected output, so one reports a failure.

That's fine. I usually update the expected output once the code has merged (that way it's "officially" part of `zppy`).

> For the complete integration tests, the status files for mpas_analysis and tc have the weird 'NING ' status.

This is documented in #242.

golaz (approved):
Approved by visual inspection.

Issue #233, opened 2022-04-07
Create example configuration files for other machines


forsyth2:
Create versions of `test_complete_run.cfg` for Cori and Compy. (Don't create new integration tests for those machines -- just create new configuration files to manually test/debug on them).

forsyth2:
Perhaps we can use Mache to specify paths in the example files. (#234 was implemented with a generic file that is used to generate the actual examples).

forsyth2:
#264, if implemented, would resolve this issue.

forsyth2:
Replaced by #264.

PR #244, opened 2022-05-16
Misc Updates

### LINKED ISSUE #238: Add slurm 'account'
In the SLURM header (https://github.com/E3SM-Project/zppy/blob/main/zppy/templates/slurm_header.sh), the slurm attribute 

`#SBATCH --account=[...]` 

is hardwired for several machines (e3sm for compy, cori). Since we may have collaborators on NERSC needing to use a different account, we should make it a parameter.

### LINKED ISSUE #239: Use `$USER` for paths in tests and docs
`git grep forsyth2` shows many results as example paths for tests and docs. It would be better to simply specify `$USER` in these places.

### LINKED ISSUE #240: Use Mache for all machine-specific paths
Use Mache (https://github.com/E3SM-Project/mache) for all machine-specific paths

forsyth2:
Misc Updates:

- Update examples code (#233/#234) so that user/machine-specific `cfg` files can be generated.
- Using Mache (https://github.com/E3SM-Project/mache). Resolves #240.
- For all paths mentioning a specific username: if in Python, use `os.environ.get("USER")` (or Mache), else write `$USER`. Resolves #239.
- Remove parameters in examples code (#233/#234) and the bundles test (#243) that were made unnecessary by #237.
- Replace paths in `dry_run` tests with simple placeholders `...`. This makes it easier to maintain the tests, by not needing to update paths.
- Change data paths to use the paths that Mache directs to.
- Improve capability to automatically determine/"guess" data paths (from #237).
- Print the URL for tasks that generate graphics.
- Make `account` a parameter. Resolves #238.


tests/integration/test_complete_run.cfg
```diff
@@ -64,17 +64,18 @@ years = "1850:1854:2",
...
 [e3sm_diags]
 active = True
 grid = '180x360_aave'
-obs_ts = '/lcrc/soft/climate/e3sm_diags_data/obs_for_e3sm_diags/time-series'
+obs_ts = '/lcrc/group/e3sm/diagnostics/observations/Atm/time-series'
 ref_final_yr = 2014
 ref_start_yr = 1985
-reference_data_path = '/lcrc/soft/climate/e3sm_diags_data/obs_for_e3sm_diags/climatology'
+# TODO: this directory is missing OMI-MLS
+reference_data_path = '/lcrc/group/e3sm/diagnostics/observations/Atm/climatology'
```
forsyth2, line 78: @chengzhuzhang This directory is missing OMI-MLS. Is that ok?

tests/integration/test_complete_run.cfg
```diff
@@ -83,18 +84,18 @@ years = "1850:1854:2", "1850:1854:4",
...
-  environment_commands="source /home/ac.forsyth2/miniconda3/etc/profile.d/conda.sh; conda activate e3sm_diags_dev_20220228"
+  # To run this test, replace conda environment with your e3sm_diags dev environment
+  environment_commands="source /home/$USER/miniconda3/etc/profile.d/conda.sh; conda activate e3sm_diags_dev_20220228"
   sets = "qbo",
   ts_subsection = "atm_monthly_180x360_aave"
 
   [[ atm_monthly_180x360_aave_tc_analysis ]]
   # Running as its own subtask because tc_analysis requires jobs to run sequentially, which slows down testing
   sets = "tc_analysis",
+  # TODO: The data from this path doesn't appear to be in /lcrc/group/e3sm/diagnostics/observations/Atm
```
forsyth2, line 98: @chengzhuzhang The data from this path doesn't appear to be in `/lcrc/group/e3sm/diagnostics/observations/Atm`. What is the proper path for the TC obs data?

zppy/templates/slurm_header.sh
```diff
@@ -16,7 +16,7 @@
 # Running on cori-haswell or cori-knl depending on 'partition'
 
 #SBATCH  --job-name={{ prefix }}
-#SBATCH  --account=e3sm
+#SBATCH  --account={{ account }}
```
forsyth2, line 19: @golaz Anvil's account is `condo`, not `e3sm`, so what should the default in `zppy/templates/default.ini` be?

Also, Chrysalis doesn't even include an account -- should it?

zppy/utils.py
```diff
@@ -59,6 +61,18 @@ def getTasks(config, section_name):
             if get_active_status(task):
                 tasks.append(task)
 
+    # Replace `$USER` from the cfg file with the actual username
+    try:
+        machine_info = MachineInfo()
+        username = machine_info.config.get("web_portal", "username")
+    except ValueError:
+        # For unit tests
+        username = os.environ.get("USER")
```
forsyth2, line 70: @xylar It seems like `os.environ.get("USER")` would always be simpler than using Mache for the username. Is there a reason it'd be better to use Mache for the username?

zppy/utils.py
```diff
@@ -59,6 +61,18 @@ def getTasks(config, section_name):
             if get_active_status(task):
                 tasks.append(task)
 
+    # Replace `$USER` from the cfg file with the actual username
```
forsyth2, line 64: Note to self: this reads like an instruction rather than a description.

forsyth2 (commented):
@golaz @chengzhuzhang @xylar This pull request is full of simple, yet useful, improvements for `zppy`. I don't think it needs a thorough review, but I have tagged each of you regarding specific questions.

Also, I would like to wait for #243 to merge before merging this, since the conflicts will be easier to handle on this branch.

tests/integration/test_complete_run.cfg
chengzhuzhang: There is a directory structure under /Atm. i.e. https://web.lcrc.anl.gov/public/e3sm/diagnostics/observations/Atm/tc-analysis/

tests/integration/test_complete_run.cfg
chengzhuzhang: The data is here; https://web.lcrc.anl.gov/public/e3sm/diagnostics/observations/Atm/climatology/OMI-MLS/

forsyth2: I see OMI-MLS in `/lcrc/group/e3sm/public_html/diagnostics/observations/Atm/climatology`. However, the `base_path` from Mache is `/lcrc/group/e3sm/public_html/diagnostic_output`, not `/lcrc/group/e3sm/public_html/diagnostics/`. So, OMI-MLS won't get picked up here. I think that's fine for a test though.

tests/integration/test_complete_run.cfg
forsyth2: Same as other comment. I see the TC-analysis in `/lcrc/group/e3sm/public_html/diagnostics/observations/Atm/tc-analysis` but the `base_path` from Mache is `/lcrc/group/e3sm/public_html/diagnostic_output`, not `/lcrc/group/e3sm/public_html/diagnostics/`. 

In this case however, since we have the `tc_obs` parameter, we can set it to be something different. That said, I think the better option would be to move the TC data to the Mache path.

tests/integration/test_last_year.py
```diff
@@ -4,6 +4,7 @@
 
 class TestLastYear(unittest.TestCase):
     def test_last_year(self):
+        # Note that this integration tests does not have any expected files in /lcrc/group/e3sm/public_html/zppy_test_resources/
```
forsyth2, line 7: `tests` -> `test`

zppy/utils.py
xylar: Oh, okay, that's fine.  There is a good reason to have it as a config option in `mache`, which is that it can then be referred to within other config options.  But I completely agree in this context.

zppy/templates/slurm_header.sh
forsyth2: Chrysalis defaults to `e3sm` so I have set it to use `e3sm` as a default.

PR #249, opened 2022-06-03
Bump to 2.1.0rc1

forsyth2:
Bump to 2.1.0rc1


forsyth2:
Note: `tbump` automatically incremented the `jinja2` version since it was also at `2.0.0`. The second commit fixes this.

PR #255, opened 2022-06-10
Fix build number

forsyth2:
Fix build number

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -10,7 +10,7 @@ source:
   git_rev: v{{ version }}
 
 build:
-  number: 1
+  number: 0
   script: "{{ PYTHON }} -m pip install . --no-deps -vv"
   noarch: python
   entry_points:


xylar:
Hmm, that's actually not the right thing to do either.  You should reset the build number to 0 only as part of incrementing to the next version number (e.g. rc3).  It doesn't matter in this case but it would not be good if you were to actually make build 0 and upload it to anaconda.org because we already had build 1, which gets higher priority.

This is one reason to move over to conda-forge when if and when you can -- they remind you every time you do a pull request to either increment the build number if the version is the same or reset it to zero if this is a new version.  But I understand that zppy can't go to conda-forge yet, and it might not make sense to ever move over there.

forsyth2:
> You should reset the build number to 0 only as part of incrementing to the next version number (e.g. rc3)

Ah, that makes sense. Luckily, I don't intend on uploading this build to anaconda.org, so at least it will be in place for rc3.

> This is one reason to move over to conda-forge

I don't know too much about conda-forge, but I can look into it.

xylar:
The only real requirement is that all your dependencies need to be on conda-forge.  I think that's the case, but I think we had thought that `zppy` was too specialized so maybe we wouldn't move it there.  But `mache` is also super specialized and it's on conda-forge so maybe it's fine.

PR #265, opened 2022-06-17
Misc Fixes

### LINKED ISSUE #262: For NERSC, ability to specify partition in bundle section. Fix typo.
Very minor issue.

When bundling on NERSC, it would be nice to be able to specify the 'partition' under the [bundle] section. Something like

[bundle]

  [[ zppy_0151_all ]]
  nodes = 1
  walltime = "48:00:00"
  partition = haswell

But this is currently not possible, since zppy will check the partition value under [default]. Maybe modify to be either under [default] or [bundle]:

https://github.com/E3SM-Project/zppy/blob/ba49505d32abcbdcc002c9786b7211eab7e300fd/zppy/__main__.py#L88-L95

Also, there is a typo on line 93 "parition".

forsyth2:
Misc Fixes, resolves #262

- Fix typo
- Enable `partition` to be set at the `bundle` level
- Fix invalid `ilamb_run` return value which could cause scripts to not run.


forsyth2:
@golaz I wanted to at least get these changes merged. 

Note that in the example `cfg` above, `environment_commands` needs to be specified explicitly. That's because the current resolution sequence is 1) In `[default]`, `partition` is set to `knl` and `environment_commands` is not set (defaults to `""`) 2) `environment_commands` gets automatically set to the `knl` Unified environment, 3) In `[bundle]`, `partition` is overridden to be `haswell` but `environment_commands` remains the same (i.e., the script for `knl`), 4) The updated `environment_commands` gets used (i.e., the one from the `cfg` with the `[bundle]` parameters merged in).

The problem with automatically determining `environment_commands` for a bundle is the following: I believe it would be difficult to determine if the value for `environment_commands` was passed in from the `cfg` at the `bundle` level (in which case we wouldn't want to "guess" the correct value), or if it was the one auto-generated in `__main__.py` (in which case, we would). I'm not sure though.

Issue #208, opened 2022-03-09
Allow variable values in `ocean_month.py`


forsyth2:
In `zppy/templates/ocean_month.py`, Lines 32-35, `10` is hard-coded:
```
for y in range(start_yr, end_yr, 10):

    year1 = y
    year2 = y + 10 - 1
```
This was causing issues addressing #128 because the tests were using fewer than 10 years.

A related issue, on line 29: `tunits = "days since 0001-01-01 00:00:00"`, which I believe is only true if we're running from year `0001` and not say `1850`.

xylar:
> A related issue, on line 29: tunits = "days since 0001-01-01 00:00:00", which I believe is only true if we're running from year 0001 and not say 1850.

I don't think this is a problem.  It might be more aesthetic to have:
``` python
tunits = f"days since {start_yr:d04}-01-01 00:00:00"
```
but your later calls to `date2num(` take `tunits` as an argument so they will work as expected regardless of the reference year.

xylar:
```python
year2 = y + 10 - 1
```
Presumably this should be the following?
``` python
year2 = min(y + 10 -1, end_yr)
```

PR #268, opened 2022-06-21
Golaz1/2.1.0 misc fixes

golaz:
Misc minor fixes found while testing E3SM Unified 1.7.0rc1 on NERSC Cori-haswell.

Currently running additional tests, but those fixes should go in before the next unified rc.

### DIFF
--- a/zppy/ilamb_run.py
+++ b/zppy/ilamb_run.py
@@ -28,7 +28,7 @@ def ilamb_run(config, scriptDir, existing_bundles):
     # --- List of ilamb_run tasks ---
     tasks = getTasks(config, "ilamb_run")
     if len(tasks) == 0:
-        return []
+        return existing_bundles
 
     # --- Generate and submit ilamb_run scripts ---
     dependencies = []
--- a/zppy/templates/ocean_month.py
+++ b/zppy/templates/ocean_month.py
@@ -29,10 +29,10 @@ tcalendar = "noleap"
 tunits = "days since 0001-01-01 00:00:00"
 
 # Loop over year sets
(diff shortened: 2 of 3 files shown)


zppy/templates/ocean_month.py
```diff
@@ -29,10 +29,10 @@
 tunits = "days since 0001-01-01 00:00:00"
 
 # Loop over year sets
-for y in range(start_yr, end_yr, 10):
+for y in range(start_yr, end_yr, ts_num_years):
 
     year1 = y
-    year2 = y + 10 - 1
+    year2 = y + ts_num_years - 1
```
forsyth2, line 35: In #219 I set `year2 = min(y + 10 - 1, end_yr)` here as was suggested on #208. I'm assuming this change here would actually be able to close out #208 then?

zppy/ilamb_run.py
```diff
@@ -28,7 +28,7 @@ def ilamb_run(config, scriptDir, existing_bundles):
     # --- List of ilamb_run tasks ---
     tasks = getTasks(config, "ilamb_run")
     if len(tasks) == 0:
-        return []
+        return existing_bundles
```
forsyth2, line 31: This was already fixed in #265.

forsyth2 (approved):
Thanks for these updates. I'll need to test myself, but from looking at the code, they look good.

zppy/templates/ocean_month.py
forsyth2: Note: the number of years still has to be divisible by the increment.

PR #306, opened 2022-09-22
Hemispheric Averaging

### LINKED ISSUE #302: Include hemispheric averaging
Include new hemispheric averaging from NCO. Add figures for hemispheric time-series to global-time-series. To keep backwards-compatibility, we’ll have to check the dimensions: if regional dimension is not there, do what we were doing before. Keep `glb` directory structure?

forsyth2:
Include hemispheric averaging in `zppy`. Resolves #302

### DIFF
--- a/tests/integration/template_complete_run.cfg
+++ b/tests/integration/template_complete_run.cfg
@@ -42,6 +42,7 @@ years = "1850:1854:2",
   vars = "PRECT"
 
   [[ atm_monthly_glb ]]
+  environment_commands = 'module purge; module load gcc/9.2.0-ugetvbp; module load openmpi/4.0.4-hpcx-hghvhj5; export PATH="~ac.zender/bin_chrysalis:${PATH}"; export PATH="~ac.zender/bin_chrysalis:~ac.zender/anaconda/bin:${PATH}"; export LD_LIBRARY_PATH="~ac.zender/lib_chrysalis:~ac.zender/anaconda/lib:${LD_LIBRARY_PATH}"; export NCO_PATH_OVERRIDE=Yes'
   frequency = "monthly"
   input_files = "eam.h0"
   input_subdir = "archive/atm/hist"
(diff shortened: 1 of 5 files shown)


zppy/templates/default.ini
```diff
@@ -89,6 +89,10 @@ ts_fmt = string(default="ts_only")
   tpd = integer(default=None)
   ts_fmt = string(default=None)
 
+  [[atm_monthly_glb]]
+  # No U
+  vars = string(default="FSNTOA,FLUT,FSNT,FLNT,FSNS,FLNS,SHFLX,QFLX,TAUX,TAUY,PRECC,PRECL,PRECSC,PRECSL,TS,TREFHT,CLDTOT,CLDHGH,CLDMED,CLDLOW")
```
forsyth2, line 94: @golaz The problem with this approach is that this automatically adds a `atm_monthly_glb` `ts` subtask, even if one isn't requested. The other alternative is to remove `U` in the `ts.py` file, but that would mean the user has no way to add `U` if desired.

tests/integration/template_complete_run.cfg
```diff
@@ -42,10 +42,13 @@ years = "1850:1854:2",
   vars = "PRECT"
 
   [[ atm_monthly_glb ]]
+  environment_commands = 'module purge; module load gcc/9.2.0-ugetvbp; module load openmpi/4.0.4-hpcx-hghvhj5; export PATH="~ac.zender/bin_chrysalis:${PATH}"; export PATH="~ac.zender/bin_chrysalis:~ac.zender/anaconda/bin:${PATH}"; export LD_LIBRARY_PATH="~ac.zender/lib_chrysalis:~ac.zender/anaconda/lib:${LD_LIBRARY_PATH}"; export NCO_PATH_OVERRIDE=Yes'
```
forsyth2, line 45: Keeping these `environment_commands` since the new `ncclimo` will be in the next Unified release anyway. So, no need to figure out how to get the latest `ncclimo` from `conda`.

forsyth2:
Notes:
- This pull request specifically removes the `U` variable from the `glb` `ts` subtask, since it is included by default in https://github.com/E3SM-Project/zppy/blob/main/zppy/templates/default.ini#L46. #331 is more generic -- the goal being to automatically remove any 3D variable.
- This pull request also fixes some issues with `account` and `machine` in `zppy/__main__.py` from #330.

Issue #319, opened 2022-10-05
Configuration file validation not working


forsyth2:
It looks like the configuration file validation (https://github.com/E3SM-Project/zppy/blob/main/zppy/__main__.py#L169) isn't working anymore. 

While working on #281, I tried removing a trailing comma and the validator passed, even though it has historically failed in such cases -- see #290. I tried setting `active = 4`, which I would have expected to fail since `active` expects a `bool` value; however validation passed and a `ValueError: Invalid value 4 for 'active'` appeared during runtime.

It appears that something has changed in the validation process.

forsyth2:
I tried going back to v2.0.0 (Feb. 2022 -- https://github.com/E3SM-Project/zppy/releases/tag/v2.0.0):
```
$ git checkout -b test-validation 00828089733cc3ad9351e4a98902bd20fce5c163
$ emacs tests/integration/test_complete_run.cfg
# Set `active = 4`, remove a trailing comma for `years`
$ pip install .
$ zppy -c tests/integration/test_complete_run.cfg
# Behavior is the same as above.
```
This occurs regardless of using a conda environment from July 2022 or from building a conda environment based on the setup of the code in Feb. 2022.

forsyth2:
#338 will at least fix the trailing comma error, so this is not urgent for the December 2022 release.

PR #320, opened 2022-10-05
Improve validation error

### LINKED ISSUE #281: Improve validation results error description
Currently, a validation results error produces output in a hard-to-understand dictionary: https://github.com/E3SM-Project/zppy/blob/main/zppy/__main__.py#L171

Suggestions for improving the error description:
- Pretty print the dictionary
- Print only the keys/values with `False` in the validation dictionary
- Print a link to the tutorial, which has examples of how values should look (but not for all parameters...)
- See if there's any way for the `Validator` to give more information on why it failed

forsyth2:
Improve validation error. Resolves #281. Inadvertently discovered #319 in the process.


tests/integration/template_complete_run.cfg
```diff
@@ -13,7 +13,7 @@ www = "#expand user_www#zppy_test_complete_run_www"
 [climo]
 active = True
 walltime = "00:30:00"
-years = "1850:1854:2", "1850:1854:4",
+years = "1850:1854:2", "1850:1854:4"
```
forsyth2, line 16: Launches the following:
```
climo_atm_monthly_180x360_aave_1850-1851
climo_atm_monthly_180x360_aave_1852-1853
climo_atm_monthly_180x360_aave_1850-1853
climo_atm_monthly_diurnal_8xdaily_180x360_aave_1850-1851
climo_atm_monthly_diurnal_8xdaily_180x360_aave_1852-1853
climo_atm_monthly_diurnal_8xdaily_180x360_aave_1850-1853
```
Both the `"1850:1854:2"` (2-year increment) jobs and the `"1850:1854:4"` (4-year increment) jobs are launched successfully. Does this mean the trailing comma is no longer a requirement of `string_list` parameters?

tests/integration/template_complete_run.cfg
```diff
@@ -25,9 +25,9 @@ years = "1850:1854:2", "1850:1854:4",
   vars = "PRECT"
 
 [ts]
-active = True
+active = 4
```
forsyth2, line 28: Causes `ValueError: Invalid value 4 for 'active'` at runtime rather than a failed configuration validation.

tests/integration/template_complete_run.cfg
```diff
@@ -25,9 +25,9 @@ years = "1850:1854:2", "1850:1854:4",
...
 walltime = "00:30:00"
-years = "1850:1854:2",
+years = "1850:1854:2"
```
forsyth2, line 30: If `active = True` above, then the error we get here is `ValueError: Error interpreting years 1`. `climo`'s multiple year sets above no longer appear to need a trailing comma. However, this error seems to indicate at least one comma must be present in a `string_list` parameter.

zppy/__main__.py
```diff
@@ -167,8 +168,12 @@ def _validate_config(config):
     validator = Validator()
 
     result = config.validate(validator)
+    print(type(result))
```
forsyth2, line 171: `result` is a `bool` now. In the past, it has historically been some sort of dictionary or dictionary-like object. For example, from a validation failure discussed in a 3/22 Slack message, we had:
```
Validation results={'e3sm_diags': True, 'e3sm_diags_vs_model': True, 'mpas_analysis': True, 'default': True, 'climo': {'qos': True, 'nodes': True, 'walltime': True, 'input_files': True, 'fre
quency': True, 'mapping_file': True, 'grid': True, 'years': False, 'exclude': True, 'vars': True, 'atm_monthly_180x360_aave': True}, 'ts': True, 'tc_analysis': True, 'amwg': True, 'global_ti
me_series': {'input_subdir': True, 'qos': True, 'nodes': True, 'walltime': True, 'color': True, 'years': True, 'ts_num_years': True, 'figstr': True, 'moc_file': True, 'experiment_name': True
, 'ts_years': False, 'climo_years': False, 'atmosphere_only': True}}
```

forsyth2 (commented):
It appears the behavior of `config.validate(validator)` has changed significantly.

forsyth2:
> It appears the behavior of `config.validate(validator)` has changed significantly.

@golaz See above comments regarding new issues with the validator. I can't find too much about the `validate` function. https://configobj.readthedocs.io/en/latest/validate.html# was last updated in 2014. https://github.com/DiffSK/configobj was last updated in 2020.

forsyth2:
I turned my branch of this into `years-fix` for #338. #281/#320 were intended to improve the error description from the validator. However, as reported in #319, the validator no longer warns of a missing comma. Therefore, it is simpler to just add the workaround in #338. 

Ideally, we would get the validator working again (#319), at which point we can improve its error message (#281/#320), so I'm leaving this pull request open, but removing it from the December 2022 project.

PR #328, opened 2022-10-20
update e3sm_u paths

chengzhuzhang:
Update e3sm-unified paths for both andes and acme1


forsyth2:
Sorry, I completely missed this request. Completed with #342/#344. Thanks though!

PR #333, opened 2022-10-27
Perlmutter support

### LINKED ISSUE #310: Port Cori-specific information to Perlmutter
Port Cori-specific information to Perlmutter -- e.g., places in the code/docs mentioning `haswell` or `knl`. Tests should pass on Perlmutter.

forsyth2:
Perlmutter support. Resolves #310.


docs/source/campaigns.rst
```diff
@@ -17,5 +17,5 @@ High-resolution analysis (``campaign = "high_res_v1"``) is only appropriate for
 (e.g., those with the oRRS18to6 ocean and sea-ice mesh).
 It will only work if you:
 
-- have access to the `bigmem nodes on Cori <https://docs.nersc.gov/systems/cori-largemem/#access-to-the-large-memory-nodes>`_
+- have access to the `bigmem nodes on NERSC Cori <https://docs.nersc.gov/systems/cori-largemem/#access-to-the-large-memory-nodes>`_
```
forsyth2, line 20: Is there an equivalent for Perlmutter?

docs/source/getting_started.rst
```diff
@@ -72,22 +70,6 @@ install the latest version in your own custom conda environment.
...
-     module load anaconda3/2019.03
-     source /share/apps/anaconda3/2019.03/etc/profile.d/conda.sh
-
-
-NERSC
------
-    ::
-
-     module load python/3.7-anaconda-2019.10
-     source /global/common/cori_cle7/software/python/3.7-anaconda-2019.10/etc/profile.d/conda.sh
```
forsyth2, line 88: I think we can remove these `conda` instructions since the details vary so widely between machines and between users. For instance, my current command to get `conda` on Cori is `source /global/homes/f/forsyth/miniconda3/etc/profile.d/conda.sh`.

forsyth2:
Currently running `zppy` for the complete-run tests. For reference, necessary steps:
```
# Load conda
$ source /global/homes/f/forsyth/miniconda3/etc/profile.d/conda.sh

# Set up conda environment
$ conda clean --all
$ conda env create -n zppy_dev_20221027 -f conda/dev.yml
$ conda activate zppy_dev_20221027
$ pip install .
$ cd ../mache
$ python -m pip install -e .
$ cd -

# Confirm Mache works with Perlmutter
$ python -c "import mache; print(mache.MachineInfo())"

# Run zppy for complete-run test
$ rm -rf /global/cfs/cdirs/e3sm/www/forsyth/zppy_test_complete_run_www/v2.LR.historical_0201
$ rm -rf /global/cfs/cdirs/e3sm/forsyth/zppy_test_complete_run_output/v2.LR.historical_0201/post
$ python tests/integration/utils.py
$ zppy -c tests/integration/generated/test_complete_run.cfg
```

docs/source/campaigns.rst
xylar: Nothing is currently listed in NERSC's documentation about large-memory nodes for Perlmutter.

docs/source/getting_started.rst
xylar: I don't recommend the procedure that was listed here so I concur with getting rid of this.  There is no added benefit that I'm aware of in using someone else's anaconda module as a starting point rather than installing your own Miniconda or Mambaforge.

zppy/templates/amwg.csh
```diff
@@ -436,7 +436,7 @@ set sig_lvl = 0.05           # level of significance
 {% if machine == 'compy' %}
 #setenv DIAG_HOME /compyfs/software/amwg/amwg_diag
 setenv DIAG_HOME /compyfs/diagnostics/amwg/amwg_diag
-{% elif machine == 'cori' %}
+{% elif machine in ['pm-cpu', 'pm-gpu'] %}
 #setenv DIAG_HOME /global/cfs/cdirs/e3sm/golaz/amwg/amwg_diag
```
xylar, line 440: Take this line out?

zppy/templates/slurm_header.sh
```diff
@@ -11,9 +11,9 @@
 #SBATCH  --qos={{ qos }}
 #SBATCH  --partition={{ partition }}
 
-{% elif machine == 'cori' %}
+{% elif machine in ['pm-cpu', 'pm-gpu'] %}
```
xylar, line 14: It should be possible to have if for the particular template arguments, rather than the machines so things could be much more compact.  See what we do in compass:
https://github.com/MPAS-Dev/compass/blob/master/compass/job/job_script.template

Here's how we call it:
https://github.com/MPAS-Dev/compass/blob/master/compass/job/__init__.py#L87-L90

xylar: The result has a bit too much whitespace, so we clean it up:
https://github.com/MPAS-Dev/compass/blob/master/compass/job/__init__.py#L101-L122

xylar (approved):
Looks great!  We'll have to see how it works once E3SM-Unified is in place to test with but it looks like what we want.

docs/source/campaigns.rst
forsyth2: So does that mean the high-res campaign won't work now? Since it apparently only runs on Cori's `bigmem` nodes?

zppy/templates/amwg.csh
forsyth2: The commented line?

zppy/templates/slurm_header.sh
forsyth2: Since this isn't required for Perlmutter support, I created #333 to address this.

zppy/templates/amwg.csh
xylar: Yes.  No useful to anyone but @golaz, right?

xylar:
> Here, I instead got Lmod has detected the following error: The following module(s) are unknown: "gcc/9.2.0-ugetvbp" -- so perhaps that module isn't available on Perlmutter?

Yeah, totally different modules on Perlmutter.  Where is this coming from?  Not from mache, I think.

xylar:
Yeah, all of this:
https://github.com/E3SM-Project/zppy/blob/932061fe020eabb5aca9f52afdf4380f9fffda91/tests/integration/template_bundles.cfg#L53
https://github.com/E3SM-Project/zppy/blob/932061fe020eabb5aca9f52afdf4380f9fffda91/tests/integration/template_bundles.cfg#L53
is manchine-specific.

forsyth2:
Will test once there is an E3SM Unified release candidate for Perlmutter.

PR #352, opened 2022-11-15
add aerosol budget tables as a default set for e3sm_diags

chengzhuzhang:
The new e3sm_diags release candidate has this new aerosol budget diagnostics set. And I added this to the default list in zppy.


forsyth2 (approved):
Thanks @chengzhuzhang!

PR #363, opened 2022-11-24
Update defaults in docs

forsyth2:
Update defaults in docs


forsyth2:
This wasn't actually necessary since no `default.ini` changes were made after #353.

PR #374, opened 2022-12-06
Fix input file error message

### LINKED ISSUE #125: Add error message when input files are not found
@golaz suggested we should add an error message in the `zppy` `ts` task when the input files are not found. Currently, `input.txt` becomes empty and `ncclimo` returns an obscure message about an incorrect number of input files for climatology (because it reverts to `climo` instead of `ts` mode).

forsyth2:
Fix input file error message. Resolves #125.

### DIFF
--- a/zppy/templates/climo.bash
+++ b/zppy/templates/climo.bash
@@ -88,6 +88,12 @@ fi
 {%- endraw %}
 
 ls {{ case }}.{{ input_files }}.????-*.nc > input.txt
+if grep -q "*" input.txt; then
+  cd {{ scriptDir }}
+  echo 'Missing input files'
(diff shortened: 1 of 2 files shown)


forsyth2:
Tested by setting `years = "1500:1502:2",` (e.g., years that would not be in the data) for `[climo]  [[ atm_monthly_diurnal_8xdaily_180x360_aave ]]` and `[ts] [[ atm_monthly_180x360_aave ]]`.

forsyth2:
The issue was that if `ls {{ case }}.{{ input_files }}.????-*.nc` does not match any files, then `input.txt` will be filled with files ending in `*.nc` rather than actual input files. Thus, we run `grep` on the `*`.

PR #389, opened 2022-12-20
Refactor `coupled_global.py`

forsyth2:
Refactor `coupled_global.py` by using a common function for all plots. This will make it easier to add new plots.

- [x] Finish converting plot functions to use the common function.
- [x] Compare `param_dict` for each plot with its unique function in the current https://github.com/E3SM-Project/zppy/blob/main/zppy/templates/coupled_global.py
- [x] Run pre-commit checks.
- [x] Run integration tests for `global_time_series`. In theory, no images will need to be updated (the generated bash files will change, however).

### DIFF
--- a/zppy/templates/coupled_global.py
+++ b/zppy/templates/coupled_global.py
@@ -132,249 +132,220 @@ def get_ylim(standard_range, extreme_values):
 # -----------------------------------------------------------------------------
 # Plotting functions
 
-# Net TOA flux (restom)
+# 1
 def plot_net_toa_flux_restom(ax, xlim, exps):
-    ax.set_xlim(xlim)
-    extreme_values = []
-    for exp in exps:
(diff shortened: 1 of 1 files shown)


zppy/templates/coupled_global.py
```diff
@@ -132,47 +132,157 @@ def get_ylim(standard_range, extreme_values):
...
+        "check_exp_year": True,
+        "default_ylim": [-1.5, 1.5],
+        "do_add_line": True,
+        "do_add_trend": True,
+        "lw": 1.0,
+        "set_axhline": True,
+        "set_legend": True,
+        "shorten_year": False,
+        "title": "Net TOA flux (restom)",
+        "var": np.array(exp["annual"]["RESTOM"]),
```
forsyth2, line 148: `"var": lambda exp: np.array(exp["annual"]["RESTOM"]),`

PR #417, opened 2023-04-06
Add support for clm2 cam in e3sm_to_cmip call

chengzhuzhang:
There is an issue reported by @BunnyVon that e3sm_to_cmip call failed during time-series generation. The issue is that the input files has old components names `clm2` and `cam`. This PR updates the e3sm_to_cmip command to support these file names.


forsyth2 (approved):
@chengzhuzhang Thanks. I haven't run the tests on this change, but it looks reasonable.

chengzhuzhang:
Thank you, Ryan. I did test a dry run and the results bash scripts look reasonable. I'm merging now.

PR #432, opened 2023-06-09
Remove AMWG task

golaz:
Removing old AMWG zppy task which is no longer supported.


tests/test_sections.cfg
```diff
@@ -28,5 +28,3 @@ active = False
 #[mpas_analysis]
 #active = False
```
forsyth2, line 30: This blank line needs to be removed to pass the `pre-commit run --all-files` checks.

forsyth2 (approved):
Just one comment. Other pre-commit checks pass and the tests look good.

PR #433, opened 2023-06-14
Update release docs

forsyth2:
Update release docs

### DIFF
--- a/docs/source/dev_guide/release.rst
+++ b/docs/source/dev_guide/release.rst
@@ -68,6 +68,14 @@ Releasing on GitHub: release candidates
 
 1. Create a tag for the release candidate at https://github.com/E3SM-Project/zppy/tags.
 
+     ::
+
+	$ git checkout main
+	$ git fetch upstream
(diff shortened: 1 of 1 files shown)


docs/source/dev_guide/release.rst
```diff
@@ -68,6 +68,14 @@ Releasing on GitHub: release candidates
 
 1. Create a tag for the release candidate at https://github.com/E3SM-Project/zppy/tags.
 
+     ::
+
+	$ git checkout main
+	$ git fetch upstream
+	$ git rebase upstream/main
```
xylar, line 75:
```
$ git reset --hard upstream/main
```
This would be better because you don't typically want to keep any changes you might have made in your local `main`.

forsyth2: Ah, I typically don't make changes on my local `main`, but I see how this command would be more fail-safe. Thanks!

forsyth2: Added as part of #440.

PR #429, opened 2023-05-24
Change workflow to mamba

forsyth2:
Change workflow to mamba. Resolves #397.

PR #434, opened 2023-06-15
Fix GitHub Actions

forsyth2:
Fix GitHub Actions


forsyth2:
@xylar You mentioned in https://github.com/E3SM-Project/zppy/pull/429#issuecomment-1562210348 "You want to avoid mixing conda-forge and pip packages whenever possible." 

However, reverting that change, as done in this PR, seems to be the only way to get the GitHub Actions checks passing. I imagine the same will hold true for the equivalent `zstash` error: https://github.com/E3SM-Project/zstash/pull/271#discussion_r1231591802

PR #438, opened 2023-06-16
Change workflows to mamba, take 2

xylar:
This merge attempts to move all workflows (including building the documentation) to `mamba` instead of `pip`.

### DIFF
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -118,10 +118,6 @@ jobs:
             ${{ runner.os }}-pip-
             ${{ runner.os }}-
 
-      # Using pip for Sphinx dependencies because it takes too long to reproduce a conda environment (~10 secs vs. 3-4 mins)
-      - name: Install Dependencies
-        run: |
-          mamba install -y sphinx==5.2.3 sphinx_rtd_theme==1.0.0 sphinx-multiversion==0.2.4 docutils==0.16
       - name: Build Sphinx Docs
         run: |
           cd docs
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -64,7 +64,7 @@ jobs:
         uses: conda-incubator/setup-miniconda@v2
         with:
           activate-environment: zppy_dev
-	  miniforge-variant: Mambaforge
+          miniforge-variant: Mambaforge
           miniforge-version: latest
           use-mamba: true
           mamba-version: "*"
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -64,6 +64,10 @@ jobs:
         uses: conda-incubator/setup-miniconda@v2
         with:
           activate-environment: zppy_dev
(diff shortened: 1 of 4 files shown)


conda/dev.yml
```diff
@@ -26,10 +26,9 @@ dependencies:
   # If versions are updated, also update in `.github/workflows/build_workflow.yml`
   # =================
   - sphinx=5.2.3
+  - sphinx-multiversion=0.2.4
```
xylar, line 29: I don't think this matters but using single rather than double `=` here to match other dependencies.

.github/workflows/build_workflow.yml
```diff
@@ -64,11 +64,13 @@ jobs:
...
           activate-environment: zppy_dev
+          miniforge-variant: Mambaforge
+          miniforge-version: latest
+          use-mamba: true
+          mamba-version: "*"
           environment-file: conda/dev.yml
           channel-priority: strict
           auto-update-conda: true
-          # IMPORTANT: This needs to be set for caching to work properly!
-          use-only-tar-bz2: true
```
xylar, line 71: I don't think this is necessary any longer.

xylar: And I think it causes problems because most newer releases are as `.conda` files.

xylar: This may explain why `sphinx-multiversion` couldn't be installed.

.github/workflows/build_workflow.yml
```diff
@@ -98,27 +100,37 @@ jobs:
...
-          # This path is specific to Ubuntu
-          path: ~/.cache/pip
-          # Look to see if there is a cache hit for the corresponding requirements file
-          key: ${{ runner.os }}-pip-publish-docs
-          restore-keys: |
-            ${{ runner.os }}-pip-
-            ${{ runner.os }}-
-
-      # Using pip for Sphinx dependencies because it takes too long to reproduce a conda environment (~10 secs vs. 3-4 mins)
-      - name: Install Dependencies
```
xylar, line 118: This should be much better with `mamba` instead of `conda`.

xylar:
@forsyth2 and @tomvothecoder, I think the problem with #429 and #434 were that you did not actually switch to creating a mambaforge environment for building the docs.  The environment was still being created with `pip` and therefore `mamba` was not installed.  An additional problem was the `use-only-tar-bz2: true` when you were creating the conda environment for testing.  This prevents newer packages like `sphinx-multiversion` from being installed and I have found that caching works even without this line.

xylar:
Unfortunately, we do not have a good way of testing these workflows until we merge this branch and make an `rc2` tag.

forsyth2 (approved):
Thanks @xylar! This looks good from what I can tell. Should I merge it?

xylar:
@forsyth2, I *think* so but I really can't be sure until we actually try a merge and see how it goes.

xylar:
@forsyth2, it seems like things worked fine so let's proceed with zstash.

PR #460, opened 2023-07-20
Skip glb-only plots for hemispheric ts

forsyth2:
Follow up to #459 -- if a plot only uses `glb` data, with no option for hemispheric plotting, then it should not be plotted on the hemispheric time series.

### DIFF
--- a/zppy/templates/coupled_global.py
+++ b/zppy/templates/coupled_global.py
@@ -133,7 +133,7 @@ def get_ylim(standard_range, extreme_values):
 # Plotting functions
 
 # 1
-def plot_net_toa_flux_restom(ax, xlim, exps):
+def plot_net_toa_flux_restom(ax, xlim, exps, rgn):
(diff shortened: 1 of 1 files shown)


forsyth2:
@golaz @chengzhuzhang I found it most straightforward to leave the glb-only plots blank on the hemispheric time series pages. 1) It is very difficult to tell at the page setup level whether a plot will only have glb-data, 2) I think this actually looks better visually, as plots will be in the same row/col across all 3 (glb, n, s) for easy comparison.

Let me know if this is acceptable output/formatting:

glb:
image

n:
image

s:
image

chengzhuzhang:
Hi Ryan, thanks for working on this. A small glitch: title of the temperature time-series plot for glb, N, S, all show global, it might be worth to fix..

forsyth2:
Good catch, thanks @chengzhuzhang

PR #466, opened 2023-07-28
Fixes for rc4

forsyth2:
Fixes for rc4


forsyth2:
The second commit undoes the work to test `e3sm_to_cmip` in its own environment. the implementation was enough for ad-hoc testing, but is not robust enough for production use. In particular, `e3sm_to_cmip_environment_commands` wasn't set to the `ts` `environment_commands` by default, meaning it would always have to be set.

A more robust solution is giving `e3sm_to_cmip` its own zppy task, as described in #467.

PR #472, opened 2023-08-04
Fix mvm paths

### LINKED ISSUE #471: For model vs model e3sm_diags runs incorrect time series path for test data
Reported by @golaz, in the qbo model vs model results, test and reference run appears using the same data (reference run data). Same applied to other time series based analysis, such as enso diags. The run is configured with zppy, I suspect the [e3sm_diags template](zppy/templates/e3sm_diags.bash) has a bug to assign test data path. 

image

Command line to reproduce:
Screen Shot 2023-08-03 at 5 28 00 PM

forsyth2:
Fix mvm paths. Resolves #471.

### DIFF
--- a/zppy/templates/e3sm_diags.bash
+++ b/zppy/templates/e3sm_diags.bash
@@ -236,7 +236,7 @@ from e3sm_diags.parameter.tc_analysis_parameter import TCAnalysisParameter
 from e3sm_diags.run import runner
 
 short_name = '${short}'
-test_ts = '${ts_dir_destination}'
+test_ts = '${ts_dir_primary}'
 start_yr = int('${Y1}')
 end_yr = int('${Y2}')
 num_years = end_yr - start_yr + 1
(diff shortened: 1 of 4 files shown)


zppy/templates/e3sm_diags.bash
```diff
@@ -236,7 +236,7 @@ from e3sm_diags.parameter.tc_analysis_parameter import TCAnalysisParameter
 from e3sm_diags.run import runner
 
 short_name = '${short}'
-test_ts = '${ts_dir_destination}'
+test_ts = '${ts_dir_primary}'
```
forsyth2, line 239: Without this change, the `mvm` test actually wasn't generating plots. Multiple errors were going unnoticed, including:
```
IndexError: list index out of range
RuntimeError: Invalid year range specified for test/reference time series data: end_year=1853>1851=var_end_yr
RuntimeError: Requested years are outside of available sst obs records.
IndexError: index 0 is out of bounds for axis 0 with size 0
```

Issue #485, opened 2023-08-18
Unified rc12 and rc14 testing: `e3sm_diags` takes too long on Chrysalis


forsyth2:
Even after setting `walltime = "4:00:00"` for the `[[ atm_monthly_180x360_aave ]]` subtask in `tests/integration/generated/test_complete_run_chrysalis.cfg`, the following jobs still run out of time:

```
e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1850-1851
e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1850-1853
e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1852-1853
```

This is a change from `walltime = "2:00:00"` in https://github.com/E3SM-Project/zppy/blob/main/tests/integration/generated/test_complete_run_chrysalis.cfg#L93

It also took significantly longer to get compute nodes with this extended time limit.

forsyth2:
@chengzhuzhang Is there any reason it would be taking so long to run E3SM Diags?

chengzhuzhang:
@forsyth2 When I tested on `perlmutter` and `compy` I don't see long time to run. `compy` is about 10% longer then `perlmutter`. I didn't tested on `chrysalis`. how about your test on the other two machines?

forsyth2:
@chengzhuzhang E3SM Diags runs in the expected amount of time on both Perlmutter and Compy.

xylar:
Could this just be because Chrysalis is being hit pretty hard right now and the disk is slow?

forsyth2:
@xylar I suppose that's a possibility. It would be good if it's nothing on our end. Unfortunately though, it is delaying testing since I need to get E3SM Diags results to run the tests on in the first place...

forsyth2:
It looks like the E3SM Diags tasks completed successfully when given between 4 and 5 hours to run.

chengzhuzhang:
Unfortunately, I can reproduce this problem with the complete zppy test. When setting 2 hours time limit for e3sm_diags runs, jobs were cancelled due to time limit.

xylar:
Do you have individual timings for different steps in `e3sm_diags`?  If so, can you compare a run with `rc9` or `rc10` to `rc12` to see what changed?

chengzhuzhang:
I did a set of e3sm_diags runs comparing timing between `rc9` and `rc12`. 
`rc9`:
real	7m33.804s
user	45m46.510s
[...]

PR #487, opened 2023-08-22
Testing fixes for Chrysalis

forsyth2:
Testing fixes for Chrysalis.


forsyth2:
Chrysalis appears to be all set for a final release. Note that #485 appears to be an issue with E3SM Diags, not zppy.

Issue #406, opened 2023-02-21
`zppy` cannot discover the hostname from a compute node on chrysalis


mahf708:
```
[ac.ngmahfouz@chr-0493 pipd_cdf6c6e]$ source /lcrc/soft/climate/e3sm-unified/load_latest_e3sm_unified_chrysalis.sh
(e3sm_unified_1.8.0_chrysalis) [ac.ngmahfouz@chr-0493 pipd_cdf6c6e]$ source /lcrc/soft/climate/e3sm-unified/load_latest_e3sm_unified_chrysalis.shD^C
(e3sm_unified_1.8.0_chrysalis) [ac.ngmahfouz@chr-0493 pipd_cdf6c6e]$ zppy -c F2010.PD.NGD_v3atm.9ac9529.chrysalis.cfg 
Configuration file validation passed.
Traceback (most recent call last):
  File "/lcrc/soft/climate/e3sm-unified/base/envs/e3sm_unified_1.8.0_chrysalis/bin/zppy", line 10, in <module>
    sys.exit(main())
  File "/lcrc/soft/climate/e3sm-unified/base/envs/e3sm_unified_1.8.0_chrysalis/lib/python3.10/site-packages/zppy/__main__.py", line 84, in main
    machine_info = MachineInfo(machine=machine)
  File "/lcrc/soft/climate/e3sm-unified/base/envs/e3sm_unified_1.8.0_chrysalis/lib/python3.10/site-packages/mache/machine_info.py", line 75, in __init__
    raise ValueError('Unable to discover machine from host name')
ValueError: Unable to discover machine from host name
```

xylar:
@mahf708, as discussed in https://github.com/E3SM-Project/mache/issues/58.  This really is a zppy problem.  zppy needs to detect the machine when you're still on a login node and set the machine manually on compute nodes.

xylar:
I don't think there's any way for mache to handle this so it's a zppy problem if this should work when you launch jobs from a compute node.

xylar:
@mahf708, if you use the development version of zppy, you should be able to set the machine as a config option and then launch the job on a compute node.  That would be my suggestion.  I don't think @forsyth2 was interested in an automated approach but one other possibility specifically in E3SM-Unified would be to use the `$E3SMU_MACHINE` variable.

xylar:
@forsyth2, could you point @mahf708 to an example config file that has the `machine` config option?

forsyth2:
[...]

PR #511, opened 2023-09-17
Fix machine discovery on compute nodes

xylar:
This merge uses an E3SM-Unified environment variable, `E3SMU_MACHINE`, to determine the machine if  this environment variable has been set (e.g. by sourcing an E3SM-Unified load script).  This makes it possible
to detect the machine on compute nodes if you are using E3SM-Unified.

This merge also cleans up a few issues with the docs that I noticed.

Partially addresses #406

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -5,7 +5,7 @@ Contributing to This Documentation
 .. highlight:: none
 
 Getting Started
-==========================
+===============
 
 This documentation is created using
 `Sphinx <http://www.sphinx-doc.org/en/stable>`_. Sphinx is an open-source tool
@@ -19,7 +19,7 @@ After merging a PR, GitHub Actions automates the documentation building process.
 It pushes the HTML build to the ``gh-pages`` branch, which is hosted on GitHub Pages.
 
 Edit Documentation
-===============================
+==================
 
 Sphinx uses `reStructuredText <http://docutils.sourceforge.net/rst.html>`_
 as its markup language. For more information on how to write documentation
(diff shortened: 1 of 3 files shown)


xylar:
@forsyth2, let me know if you'd like me to make a separate PR for the docs changes.

xylar:
> Why only partially? This seems like a full resolution.

This only fixes the problem if someone is using E3SMU, not if you use a development env.

chengzhuzhang:
@xylar e3sm_diags has this same challenge that when use a development env, and run on compute node, it won't auto detect the machine and use machine specific info. I'm wondering if it is possible to  update `mache` for a full solution?

xylar:
@chengzhuzhang, I'm sorry but I don't know of any solution to this.  The compute nodes have super generic names and there is not a consistent way to detect what machine you are on using things like environment variables or files.  If you have a suggestion, I'm up for it.  But otherwise we simply need users to identify with a config option or command-line flag what machine they are on when they launch on compute nodes.

chengzhuzhang:
I was thinking for the machines that we support, if we can find some pattern of compute nodes names and filter them in `mache`. But if there is not a pattern, i guess asking users to provide a machine name is not too bad.

xylar:
@chengzhuzhang, on Chrysalis, it might work since the login nodes are named something like `chr-0472`.  I can add that.

But on Anvil, they are named: `b677`, which seems way too generic to use, but maybe folks feel like it's fine for our limit number of to assume b and 3 numbers is Anvil?

Compy is the same story: `n0432`.  Again is this safe to assume that anything like n and 4 numbers is a Compy compute node?  I have been hesitant.

On Perlmutter, even the login nodes are too generic so we have to fall back to an environment variable, and that probably works on compute nodes, though I can't say I've checked.  

Frontier was similar with the added problem that there was no definitive environment variable to use so I had to fall back on an environment variable related to modules.  Again, that likely works on compute nodes.

Other supported machines like Chicoma, Andes and ACME1, I don't know what the deal is there since I don't use the that much.

xylar:
@chengzhuzhang can you open this issue on https://github.com/E3SM-Project/mache/issues? This is not the right place to discuss this further.

PR #513, opened 2023-09-22
Update Github templates

forsyth2:
Update Github templates. Follow-up to #480. Resolves #430.

PR #514, opened 2023-09-25
Update Github templates

### LINKED ISSUE #430: Create issue/discussion templates
Create issue/discussion templates so that when users create new issues or discussions, they will be guided on what information to provide. Examples:

https://github.com/xCDAT/xcdat/issues/new/choose
https://github.com/xCDAT/xcdat/discussions/new?category=2-questions

forsyth2:
Update Github templates. Follow-up to #513. Resolves #430.

### DIFF
--- a/.github/DISCUSSION_TEMPLATE/questions.yml
+++ b/.github/DISCUSSION_TEMPLATE/questions.yml
@@ -39,7 +39,7 @@ body:
       attributes:
           label: What machine were you running on?
           description: |
-	      List the machine(s) you encounter the issue on (e.g., Chrysalis, Compy, Perlmutter).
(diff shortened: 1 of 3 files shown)


forsyth2:
Unfortunately, it looks like the syntax errors (e.g., the error below) only appear on merge, as opposed to each push to a branch. Hopefully, this is the last template updated needed.

Validation error

Following guidance from https://stackoverflow.com/questions/19456518/invalid-command-code-despite-escaping-periods-using-sed, we can create a `bash` function:
```
replace()
{
  find ./ -type f -exec sed -i '' -e "s/$1/$2/" {} \;
}
```
And then run:
```
replace "\t" "        "
```
(convert tabs to 8 spaces)

PR #519, opened 2023-10-03
Replace CDAT

### LINKED ISSUE #80: Explore replacement of cdscan
Recent e3sm-unified testing suggested that additional patch is needed on conda-forge for`cdscan`(which is a component of `cdat`). It would be nice to come up with a replacement. `cdscan` is used to concatenate files before running `e3sm-diags` and global mean time-series. One possibility is to use xarray's IO to read-in multiple files.

### LINKED ISSUE #346: Replace CDAT
https://github.com/conda-forge/cdtime-feedstock/pull/52 mentions that we need to replace `CDAT` in `zppy`. Note that we already have #80, which is specifically about `cdscan`.

forsyth2:
Replace CDAT. Resolves #346. Resolves #80.

### DIFF
--- a/zppy/templates/coupled_global.py
+++ b/zppy/templates/coupled_global.py
@@ -3,12 +3,15 @@ import glob
 import math
 import sys
 import traceback
-from typing import Any, List, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
+import cftime
 import matplotlib as mpl
 import matplotlib.backends.backend_pdf
 import matplotlib.pyplot as plt
 import numpy as np
+import xarray
+import xcdat  # noqa: F401
 from netCDF4 import Dataset
 from readTS import TS
 
@@ -549,11 +552,20 @@ def param_get_list(param_value):
         return param_value.split(",")
 
 
-def set_var(exp, exp_key, var_list, valid_vars, invalid_vars, rgn):
+def set_var(
+    exp: Dict[str, Any],
+    exp_key: str,
+    var_list: List[str],
+    valid_vars: List[str],
+    invalid_vars: List[str],
+    rgn: str,
+) -> None:
     if exp[exp_key] is not None:
         ts = TS(exp[exp_key])
         for var in var_list:
             try:
+                v: xarray.core.dataarray.DataArray
+                units: Optional[str]
                 v, units = ts.globalAnnual(var)
                 valid_vars.append(str(var))
             except Exception as e:
@@ -561,7 +573,7 @@ def set_var(exp, exp_key, var_list, valid_vars, invalid_vars, rgn):
                 print(f"globalAnnual failed. Invalid var = {var}")
                 invalid_vars.append(str(var))
                 continue
-            if len(v.shape) > 1:
+            if v.sizes["rgn"] > 1:
                 # number of years x 3 regions = v.shape
                 # 3 regions = global, northern hemisphere, southern hemisphere
                 # We get here if we used the updated `ts` task
@@ -574,18 +586,17 @@ def set_var(exp, exp_key, var_list, valid_vars, invalid_vars, rgn):
                     n = 2
                 else:
                     raise RuntimeError(f"Invalid rgn={rgn}")
-                v = v[:, n]  # Just use nth column
+                v = v.isel(rgn=n)  # Just use nth column
             elif rgn != "glb":
                 # v only has one dimension -- glb.
                 # Therefore it is not possible to get n or s plots.
                 raise RuntimeError(
                     f"var={var} only has global data. Cannot process rgn={rgn}"
                 )
-            exp["annual"][var] = v
             exp["annual"][var] = (v, units)
             if "year" not in exp["annual"]:
-                time = v.getTime()
-                exp["annual"]["year"] = [x.year for x in time.asComponentTime()]
+                years: np.ndarray[cftime.DatetimeNoLeap] = v.coords["time"].values
+                exp["annual"]["year"] = [x.year for x in years]
         del ts
 
 
@@ -736,7 +747,7 @@ def run(parameters, rgn):  # noqa: C901
         or ("change_sea_level" in plots_original)
     )
     use_ocn = plots_ocn or (not atmosphere_only and has_original_ocn_plots)
-    exps = [
+    exps: List[Dict[str, Any]] = [
         {
             "atmos": None
             if not use_atmos
@@ -767,7 +778,7 @@ def run(parameters, rgn):  # noqa: C901
     invalid_vars: List[str] = []
 
     # Read data
-    exp: Any
+    exp: Dict[str, Any]
     for exp in exps:
         exp["annual"] = {}
 
@@ -845,5 +856,27 @@ def run_by_region(parameters):
         run(parameters, rgn)
 
 
+def debug():
+    args = [
+        "coupled_global.py",
+        "/lcrc/group/e3sm/ac.forsyth2/zppy_test_debug_output/test-346/v2.LR.historical_0201",
+        "v2.LR.historical_0201",
+        "v2_historical_0201",
+        "1850",
+        "1860",
+        "Blue",
+        "5",
+        "net_toa_flux_restom,global_surface_air_temperature,toa_radiation,net_atm_energy_imbalance,change_ohc,max_moc,change_sea_level,net_atm_water_imbalance",
+        "False",
+        "None",
+        "None",
+        "LAISHA,LAISUN",
+        "None",
+        "glb,n,s",
+    ]
+    run_by_region(args)
+
+
 if __name__ == "__main__":
     run_by_region(sys.argv)
+    # debug()
--- a/zppy/templates/coupled_global.py
+++ b/zppy/templates/coupled_global.py
@@ -11,7 +11,6 @@ import matplotlib.backends.backend_pdf
 import matplotlib.pyplot as plt
 import numpy as np
 import xarray
-import xcdat  # noqa: F401
 from netCDF4 import Dataset
 from readTS import TS
 
@@ -586,7 +585,7 @@ def set_var(
                     n = 2
                 else:
                     raise RuntimeError(f"Invalid rgn={rgn}")
-                v = v.isel(rgn=n)  # Just use nth column
+                v = v.isel(rgn=n)  # Just use nth region
             elif rgn != "glb":
                 # v only has one dimension -- glb.
                 # Therefore it is not possible to get n or s plots.
@@ -749,24 +748,24 @@ def run(parameters, rgn):  # noqa: C901
     use_ocn = plots_ocn or (not atmosphere_only and has_original_ocn_plots)
     exps: List[Dict[str, Any]] = [
         {
-            "atmos": None
-            if not use_atmos
-            else "{}/post/atm/glb/ts/monthly/{}yr/".format(case_dir, ts_num_years),
-            "ice": None
-            if not plots_ice
-            else "{}/post/ice/glb/ts/monthly/{}yr/".format(case_dir, ts_num_years),
-            "land": None
-            if not plots_lnd
-            else "{}/post/lnd/glb/ts/monthly/{}yr/".format(case_dir, ts_num_years),
-            "ocean": None
-            if not use_ocn
-            else "{}/post/ocn/glb/ts/monthly/{}yr/".format(case_dir, ts_num_years),
-            "moc": None
-            if not use_ocn
-            else "{}/post/ocn/glb/ts/monthly/{}yr/".format(case_dir, ts_num_years),
-            "vol": None
-            if not use_ocn
-            else "{}/post/ocn/glb/ts/monthly/{}yr/".format(case_dir, ts_num_years),
+            "atmos": f"{case_dir}/post/atm/glb/ts/monthly/{ts_num_years}yr/"
+            if use_atmos
+            else None,
+            "ice": f"{case_dir}/post/ice/glb/ts/monthly/{ts_num_years}yr/"
+            if plots_ice
+            else None,
+            "land": f"{case_dir}/post/lnd/glb/ts/monthly/{ts_num_years}yr/"
+            if plots_lnd
+            else None,
+            "ocean": f"{case_dir}/post/ocn/glb/ts/monthly/{ts_num_years}yr/"
+            if use_ocn
+            else None,
+            "moc": f"{case_dir}/post/ocn/glb/ts/monthly/{ts_num_years}yr/"
+            if use_ocn
+            else None,
+            "vol": f"{case_dir}/post/ocn/glb/ts/monthly/{ts_num_years}yr/"
+            if use_ocn
+            else None,
             "name": experiment_name,
             "yoffset": 0.0,
             "yr": ([year1, year2],),
@@ -856,27 +855,5 @@ def run_by_region(parameters):
         run(parameters, rgn)
 
 
-def debug():
-    args = [
-        "coupled_global.py",
-        "/lcrc/group/e3sm/ac.forsyth2/zppy_test_debug_output/test-346/v2.LR.historical_0201",
-        "v2.LR.historical_0201",
-        "v2_historical_0201",
-        "1850",
-        "1860",
-        "Blue",
-        "5",
-        "net_toa_flux_restom,global_surface_air_temperature,toa_radiation,net_atm_energy_imbalance,change_ohc,max_moc,change_sea_level,net_atm_water_imbalance",
-        "False",
-        "None",
-        "None",
-        "LAISHA,LAISUN",
-        "None",
-        "glb,n,s",
-    ]
-    run_by_region(args)
-
-
 if __name__ == "__main__":
     run_by_region(sys.argv)
-    # debug()
--- a/zppy/templates/coupled_global.py
+++ b/zppy/templates/coupled_global.py
@@ -740,32 +740,22 @@ def run(parameters, rgn):  # noqa: C901
         {
             "atmos": None
             if not use_atmos
-            else "{}/post/atm/glb/ts/monthly/{}yr/glb.xml".format(
-                case_dir, ts_num_years
-            ),
+            else "{}/post/atm/glb/ts/monthly/{}yr/".format(case_dir, ts_num_years),
             "ice": None
             if not plots_ice
-            else "{}/post/ice/glb/ts/monthly/{}yr/glb.xml".format(
-                case_dir, ts_num_years
-            ),
+            else "{}/post/ice/glb/ts/monthly/{}yr/".format(case_dir, ts_num_years),
             "land": None
             if not plots_lnd
-            else "{}/post/lnd/glb/ts/monthly/{}yr/glb.xml".format(
-                case_dir, ts_num_years
-            ),
+            else "{}/post/lnd/glb/ts/monthly/{}yr/".format(case_dir, ts_num_years),
             "ocean": None
             if not use_ocn
-            else "{}/post/ocn/glb/ts/monthly/{}yr/glb.xml".format(
-                case_dir, ts_num_years
-            ),
+            else "{}/post/ocn/glb/ts/monthly/{}yr/".format(case_dir, ts_num_years),
             "moc": None
             if not use_ocn
             else "{}/post/ocn/glb/ts/monthly/{}yr/".format(case_dir, ts_num_years),
             "vol": None
             if not use_ocn
-            else "{}/post/ocn/glb/ts/monthly/{}yr/glb.xml".format(
-                case_dir, ts_num_years
-            ),
+            else "{}/post/ocn/glb/ts/monthly/{}yr/".format(case_dir, ts_num_years),
             "name": experiment_name,
             "yoffset": 0.0,
             "yr": ([year1, year2],),
--- a/zppy/templates/global_time_series.bash
+++ b/zppy/templates/global_time_series.bash
@@ -42,7 +42,7 @@ use_atm={{ use_atm }}
 if [[ ${use_atm,,} == "true" ]]; then
     echo 'Create xml files for atm'
     cd ${case_dir}/post/atm/glb/ts/monthly/${ts_num_years}yr
-    cdscan -x glb.xml *.nc
+    zppy_cdscan_replacement glb.xml *.nc
     if [ $? != 0 ]; then
       cd {{ scriptDir }}
       echo 'ERROR (1)' > {{ prefix }}.status
@@ -54,7 +54,7 @@ use_lnd={{ use_lnd }}
 if [[ ${use_lnd,,} == "true" ]]; then
     echo 'Create xml files for lnd'
     cd ${case_dir}/post/lnd/glb/ts/monthly/${ts_num_years}yr
-    cdscan -x glb.xml *.nc
+    zppy_cdscan_replacement -x glb.xml *.nc
     if [ $? != 0 ]; then
       cd {{ scriptDir }}
       echo 'ERROR (2)' > {{ prefix }}.status
@@ -78,7 +78,7 @@ if [[ ${use_ocn,,} == "true" ]]; then
     echo 'Create xml for for ocn'
     export CDMS_NO_MPI=true
     cd ${case_dir}/post/ocn/glb/ts/monthly/${ts_num_years}yr
-    cdscan -x glb.xml mpaso.glb*.nc
+    zppy_cdscan_replacement glb.xml mpaso.glb*.nc
     if [ $? != 0 ]; then
       cd {{ scriptDir }}
       echo 'ERROR (4)' > {{ prefix }}.status
--- a/zppy/templates/global_time_series.bash
+++ b/zppy/templates/global_time_series.bash
@@ -39,28 +39,7 @@ global_ts_dir={{ global_time_series_dir }}
 export CDMS_NO_MPI=true
 
 use_atm={{ use_atm }}
-if [[ ${use_atm,,} == "true" ]]; then
-    echo 'Create xml files for atm'
-    cd ${case_dir}/post/atm/glb/ts/monthly/${ts_num_years}yr
-    zppy_cdscan_replacement glb.xml *.nc
-    if [ $? != 0 ]; then
-      cd {{ scriptDir }}
-      echo 'ERROR (1)' > {{ prefix }}.status
-      exit 1
-    fi
-fi
-
 use_lnd={{ use_lnd }}
-if [[ ${use_lnd,,} == "true" ]]; then
-    echo 'Create xml files for lnd'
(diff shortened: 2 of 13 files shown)


zppy/templates/global_time_series.bash
```diff
@@ -43,7 +43,7 @@ atmosphere_only=${atmosphere_only,,}
 echo 'Create xml files for atm'
 export CDMS_NO_MPI=true
 cd ${case_dir}/post/atm/glb/ts/monthly/${ts_num_years}yr
-cdscan -x glb.xml *.nc
+python cdscan_replacement.py glb.xml *.nc
```
forsyth2, line 46: `zppy` isn't finding the `cdscan_replacement.py` file. 

`global_time_series` adds Python scripts to the current directory via the block at https://github.com/E3SM-Project/zppy/blob/main/zppy/global_time_series.py#L55:
```
            c["global_time_series_dir"] = os.path.join(
                scriptDir, "{}_dir".format(prefix)
            )
            if not os.path.exists(c["global_time_series_dir"]):
                os.mkdir(c["global_time_series_dir"])
            scripts = ["coupled_global.py", "readTS.py", "ocean_month.py"]
            for script in scripts:
                script_template = templateEnv.get_template(script)
                script_file = os.path.join(c["global_time_series_dir"], script)
                with open(script_file, "w") as f:
                    f.write(script_template.render(**c))
                makeExecutable(script_file)
```

Would it make sense to do something similar for `cdscan_replacement`? It seems like there should be a simpler way since there are no template parameters in that file.

xylar: @forsyth2, why not make this a proper entry point in the `zppy` package?  That's the "right" way to do this:
https://packaging.python.org/en/latest/specifications/entry-points/

xylar: To do that, you would add your functions along with a `main()` function equivalent to your current `if __name__ == "__main__":` block to zppy somewhere and then you would edit your `setup.py` to add another entry point like this one:
https://github.com/E3SM-Project/zppy/blob/main/setup.py#L33

forsyth2: @xylar Thanks, I tried implementing it that way, but I'm getting `zppy_cdscan_replacement: command not found` errors. Yet the command does show up when I run `which` in the dev environment.
```
$ which zppy
~/miniconda3/envs/zppy_dev_issue_346/bin/zppy
$ which zppy_cdscan_replacement
~/miniconda3/envs/zppy_dev_issue_346/bin/zppy_cdscan_replacement
```

xylar: Are you using a different environment (e.g. E3SM-Unified) when you try to call `zppy_cdscan_replacement`?  If that command is part of the environment you use to launch jobs but then it's not available in the environment that's actually used when you run the jobs, it would make sense that you're getting these errors.

forsyth2: Oh, yes that's exactly what's happening. Thanks for pointing that out!!!

forsyth2:
Following the example of https://github.com/E3SM-Project/e3sm_diags/discussions/803:

Running:
```
directory=zppy
cd ${directory}

# https://anaconda.org/cdat/repo?sort=_name&sort_order=asc
for cdat_package in cd77 cdat cdat_compute_graph cdat_info cdat_jupyter_vcdat cdms2 cdp cdtime cdutil cibots compute_graph django-allow-cidr dv3d e3sm_nex esgf-compute-api esgf-compute-cert esgf-search ezget genutil image-compare jupyter-vcdat jupyterlab-git lats libcdms lmoments mesalib myproxyclient output_viewer selenium_testlib testsrunner thermo toolchain vcdat vcs vcs-js vcsaddons vtk vtk-cdat wk xmgrace
do
    git grep -n ${cdat_package} ${directory}/* > /dev/null
    if [ $? == 0 ]; then
	echo ""
	echo "git grep -n ${cdat_package} ${directory}/*"
	git grep -n ${cdat_package} ${directory}/*
    fi
done
```
Gives:
```
zppy/templates/readTS.py:1:import cdms2
zppy/templates/readTS.py:10:        self.f = cdms2.open(filename)

git grep -n cdutil zppy/*
zppy/templates/readTS.py:2:import cdutil
zppy/templates/readTS.py:69:            v = cdutil.YEAR(v)

git grep -n wk zppy/*
zppy/templates/mpas_analysis.bash:301:size=`wc -c ${identifier}/logs/taskProgress.log | awk '{print $1}'`
```

zppy/templates/global_time_series.bash
forsyth2: @xylar I'm trying to come back to this PR, but I don't think I found a way to actually call `zppy_cdcsan_replacement` in the various environments the launched jobs run in. The only two options I can think of would be:
1. Find some way to install that function into what ever environment is in place (`conda install`?)
2. Just have a Python script in the `templates` directory that can be called (as `coupled_global.py` is called for the `global_time_series` task)

xylar: @forsyth2, I think one option would be to move `zppy_cdscan_replacement` to a different package that will be in Unified and that can be installed into test conda environments.  I don't think zppy is really the right place for adding new tools precisely because it isn't necessarily actually installed in the environments used in the jobs it launches.

One possibility would be to add it to my e3sm-tools repo:
https://github.com/xylar/e3sm-tools

I could live with your option 2 above -- putting the script in `templates`.  That's probably the easiest solution but it's not very clean.  If you can come up with another place the tool can live, I would prefer that.

xylar: @forsyth2, here would be my proposed solution:
https://github.com/xylar/e3sm-tools/pull/1/files

This would require several things to happen that are needed in any case:
1. move `e3sm-tools` from my fork to E3SM-Project
2. update the conda recipe to no require git submodule
3. add `e3sm-tools` to conda-forge

Then, you could make `e3sm-tools` be a dependency of `zppy`.

If you are okay with that solution, I can make a test package for you to install while I work on those 3 steps above.

xylar: @forsyth2, I'm starting to change my mind about `e3sm-tools`.  I will probably get rid of that repo entirely and just include the pieces in a variety of conda-forge recipes for individual tools or smaller collections.  That would leave `cdscan_replacement` without a good home.

As an alternative, would it be plausible to require `zppy` to also be installed in the environment activated in the `zppy` jobs?  Then , `zppy_cdscan_replacement` would be available.  It might be good to check that the same version of `zppy` is used to launch jobs as in the environment within the jobs (e.g. by setting and then checking an environment variable).

forsyth2: @xylar Thanks for presenting some options, I'll have to think about what the best design is.

@tomvothecoder This is the discussion thread I was talking about in our meeting. Could the `cdscan` replacement just be an `xarray` call? The relevant code block is:
```
def zppy_cdscan_replacement():
    output_file = sys.argv[1]
    input_files = sys.argv[2:]
    if len(input_files) == 1:
        files = []
        # Convert text file of files to a list of files.
        with open(input_files[0]) as f:
            for line in f:
                files.append(line.strip())
        input_files = files
    dataframe = pandas.read_table(input_files)
    dataframe.to_xml(output_file)
```

xylar: It does look a lot like a complicated way of getting something like `xarray.open_mfdataset()`.  So maybe it's not needed with `xcdat`?

forsyth2: Hmm I don't think that worked but I can't seem to find any documentation as to why.

I do see on Oct. 3, Tom noted "I suggest exploring [xarray.open_mfdataset()](https://docs.xarray.dev/en/stable/generated/xarray.open_mfdataset.html)" in https://github.com/E3SM-Project/zppy/issues/346#issuecomment-1745351344, but then in this thread from Oct. 5, I'm trying to use `cdscan_replacement.py`. That seems to imply something about `open_mfdataset` wasn't working properly.

forsyth2:
@tomvothecoder On the wider topic of any CDAT usage in `zppy`, it seems like there is very little direct CDAT dependency. It appears _all_ CDAT usage in `zppy` is done in the package environments as opposed to the environment zppy itself is run from. The `zppy` dev environment doesn't even seem to depend on CDAT at all -- notice https://github.com/E3SM-Project/zppy/blob/main/conda/dev.yml doesn't appear to include any CDAT packages.

zppy/templates/global_time_series.bash
xylar: Oh, interesting! Still, I think it's worth trying again with open_mfdataset and reminding yourself what the error was. Maybe we can solve it.

forsyth2: I'm looking into it; I want to say it was something about it needing to convert to xml.

It also occurs to me that even if we use `xarray`, we'd need to call it from a `bash` script, so the same general problem (of needing to get a bash wrapper defined in the environment) would remain.

tomvothecoder: Can you explain the step-by-step requirements/description for the functionality involving `cdscan`? This would make it easier to identify the equivalent APIs needed from other packages to replace the `cdscan` code.

From my guess, in your comment [here](https://github.com/E3SM-Project/zppy/issues/346#issuecomment-1344805061), it just looks like the code takes a glob of `.nc` filepaths and converts it to XML using `cdscan`. I assume this XML is used downstream with `cdms2` (or another CDAT package).

```bash
$ git grep -n cdscan
docs/source/dev_guide/new_diags_set.rst:100:        cdscan -x ${xml_name} ${rofDir}/${v}_*.nc
zppy/templates/e3sm_diags.bash:104:        # Add this time series file to the list of files for cdscan to use
zppy/templates/e3sm_diags.bash:111:    cdscan -x ${xml_name} -f ${v}_files.txt
zppy/templates/e3sm_diags.bash:132:  cdscan -x ${xml_name} ${ts_rof_dir_source}/${v}_*.nc
zppy/templates/global_time_series.bash:46:cdscan -x glb.xml *.nc
zppy/templates/global_time_series.bash:70:    cdscan -x glb.xml mpaso.glb*.nc

$ git grep -n cdms
zppy/templates/readTS.py:1:import cdms2
zppy/templates/readTS.py:10:        self.f = cdms2.open(filename)

```

If that's the case, you can probably bypass the XML step entirely and just pass the glob paths (e.g., `${rofDir}/${v}_*.nc` directly to `xarray.open_mfdataset()`.  **However, you'll need to do some prototyping to validate a solution (I'm just throwing things out).**
```python
path =  "${rofDir}/${v}_*.nc"

dataset = xarray.open_mfdataset(path)
```

forsyth2: > Can you explain the step-by-step requirements/description for the functionality involving `cdscan`?

For `e3sm_diags`: `cdscan` generates an `xml` file, placing it in the relevant directory path parameter for `e3sm_diags` to use.

<details>
<summary> The relevant code steps </summary>

1.  `cdscan` is used to generate a `xml` file based on `nc` files corresponding to a particular variable between January of the start year and December of the end year.
```
$ git grep -n -B 2 cdscan e3sm_diags.bash
e3sm_diags.bash-102-      for file in ${ts_dir_source}/${v}_${YYYY}*.nc
e3sm_diags.bash-103-      do
e3sm_diags.bash:104:        # Add this time series file to the list of files for cdscan to use
--
e3sm_diags.bash-109-    xml_name=${v}_${begin_year}01_${end_year}12.xml
e3sm_diags.bash-110-    export CDMS_NO_MPI=true
e3sm_diags.bash:111:    zppy_cdscan_replacement ${xml_name} ${v}_files.txt
--
e3sm_diags.bash-130-  v="RIVER_DISCHARGE_OVER_LAND_LIQ"
e3sm_diags.bash-131-  xml_name=${v}_${begin_year}01_${end_year}12.xml
e3sm_diags.bash:132:  zppy_cdscan_replacement ${xml_name} ${ts_rof_dir_source}/${v}_*.nc
```

2. This is done inside the `create_links_ts` function (and also the `create_links_ts_rof` function):
```
$ git grep -n -A 3 "create_links_ts(" e3sm_diags.bash
e3sm_diags.bash:84:create_links_ts()
e3sm_diags.bash-85-{
e3sm_diags.bash-86-  ts_dir_source=$1
e3sm_diags.bash-87-  ts_dir_destination=$2
```
3. These 2 functions have created the `xml` files in the passed-in _destination_ directories:
```
$ git grep -n -B 1 "create_links_ts" e3sm_diags.bash
e3sm_diags.bash-83-
e3sm_diags.bash:84:create_links_ts()
--
e3sm_diags.bash-120-
e3sm_diags.bash:121:create_links_ts_rof()
--
e3sm_diags.bash-201-ts_dir_source={{ output }}/post/atm/{{ grid }}/ts/monthly/{{ '%dyr' % (ts_num_years) }}
e3sm_diags.bash:202:create_links_ts ${ts_dir_source} ${ts_dir_primary} ${Y1} ${Y2} 5
--
e3sm_diags.bash-205-ts_dir_ref=ts_ref
e3sm_diags.bash:206:create_links_ts ${ts_dir_source} ${ts_dir_ref} ${ref_Y1} ${ref_Y2} 6
--
e3sm_diags.bash-221-ts_rof_dir_source="{{ output }}/post/rof/native/ts/monthly/{{ ts_num_years }}yr"
e3sm_diags.bash:222:create_links_ts_rof ${ts_rof_dir_source} ${ts_rof_dir_primary} ${Y1} ${Y2} 7
--
e3sm_diags.bash-225-ts_rof_dir_ref=ts_rof_ref
e3sm_diags.bash:226:create_links_ts_rof ${ts_rof_dir_source} ${ts_rof_dir_ref} ${ref_Y1} ${ref_Y2} 8
```
4. Following one of the destination directories, we can see that `ts_dir_primary` is used to define `test_ts` in the generated `e3sm_diags` python file.
```
$ git grep -n -B 1 ts_dir_primary e3sm_diags.bash
e3sm_diags.bash-195-{% if run_type == "model_vs_obs" %}
e3sm_diags.bash:196:ts_dir_primary=ts
e3sm_diags.bash-197-{% elif run_type == "model_vs_model" %}
e3sm_diags.bash:198:ts_dir_primary=ts_test
--
e3sm_diags.bash-201-ts_dir_source={{ output }}/post/atm/{{ grid }}/ts/monthly/{{ '%dyr' % (ts_num_years) }}
e3sm_diags.bash:202:create_links_ts ${ts_dir_source} ${ts_dir_primary} ${Y1} ${Y2} 5
--
e3sm_diags.bash-278-short_name = '${short}'
e3sm_diags.bash:279:test_ts = '${ts_dir_primary}'
```
5. That `test_ts` value is used to defined the `test_data_path` for several parameters.
```
$ git grep -n test_ts e3sm_diags.bash 
e3sm_diags.bash:279:test_ts = '${ts_dir_primary}'
e3sm_diags.bash:350:enso_param.test_data_path = test_ts
e3sm_diags.bash:406:qbo_param.test_data_path = test_ts
e3sm_diags.bash:436:ts_param.test_data_path = test_ts
```

</details>

For `global_time_series`: `cdscan` generates `glb.xml` files which are ultimately opened and processed to generate a global annual average for each variable contained.

<details>
<summary> The relevant code steps </summary>

1. `cdscan` is used to generate a `glb.xml` file based on _all_ the `nc` files in a particular `glb` subdirectory, for multiple subdirectories:
```
$ git grep -n -B 1 cdscan global_time_series.bash 
global_time_series.bash-44-    cd ${case_dir}/post/atm/glb/ts/monthly/${ts_num_years}yr
global_time_series.bash:45:    zppy_cdscan_replacement glb.xml *.nc
--
global_time_series.bash-56-    cd ${case_dir}/post/lnd/glb/ts/monthly/${ts_num_years}yr
global_time_series.bash:57:    zppy_cdscan_replacement -x glb.xml *.nc
--
global_time_series.bash-80-    cd ${case_dir}/post/ocn/glb/ts/monthly/${ts_num_years}yr
global_time_series.bash:81:    zppy_cdscan_replacement glb.xml mpaso.glb*.nc
```
2. Which are then used to define values for keys in the `exp` directory:
```
$ git grep -n -B 2 "glb.xml" coupled_global.py
coupled_global.py-741-            "atmos": None
coupled_global.py-742-            if not use_atmos
coupled_global.py:743:            else "{}/post/atm/glb/ts/monthly/{}yr/glb.xml".format(
--
coupled_global.py-746-            "ice": None
coupled_global.py-747-            if not plots_ice
coupled_global.py:748:            else "{}/post/ice/glb/ts/monthly/{}yr/glb.xml".format(
--
coupled_global.py-751-            "land": None
coupled_global.py-752-            if not plots_lnd
coupled_global.py:753:            else "{}/post/lnd/glb/ts/monthly/{}yr/glb.xml".format(
--
coupled_global.py-756-            "ocean": None
coupled_global.py-757-            if not use_ocn
coupled_global.py:758:            else "{}/post/ocn/glb/ts/monthly/{}yr/glb.xml".format(
--
coupled_global.py-764-            "vol": None
coupled_global.py-765-            if not use_ocn
coupled_global.py:766:            else "{}/post/ocn/glb/ts/monthly/{}yr/glb.xml".format(
```
3. These `glb.xml` files are then passed into `set_var`, implicitly as `exp[exp_key]`
```
$ git grep -n "set_var" coupled_global.py
coupled_global.py:552:def set_var(exp, exp_key, var_list, valid_vars, invalid_vars, rgn):
coupled_global.py:786:        set_var(exp, "atmos", vars_original, valid_vars, invalid_vars, rgn)
coupled_global.py:787:        set_var(exp, "atmos", plots_atm, valid_vars, invalid_vars, rgn)
coupled_global.py:788:        set_var(exp, "ice", plots_ice, valid_vars, invalid_vars, rgn)
coupled_global.py:789:        set_var(exp, "land", plots_lnd, valid_vars, invalid_vars, rgn)
coupled_global.py:790:        set_var(exp, "ocean", plots_ocn, valid_vars, invalid_vars, rgn)
```
4. They're then passed to the `TS` class constructor:
```
$ git grep -n "exp\[exp_key\]" coupled_global.py
coupled_global.py:553:    if exp[exp_key] is not None:
coupled_global.py:554:        ts = TS(exp[exp_key])
```
5. The `glb.xml` files are ultimately opened, formerly by `cdms2.open` and with this PR by `xcdat.open_dataset`
```
$ git grep -n -A 5 "TS" readTS.py
readTS.py:4:class TS(object):
readTS.py-5-    def __init__(self, filename):
readTS.py-6-
readTS.py-7-        self.filename = filename
readTS.py-8-
readTS.py-9-        self.f = xcdat.open_dataset(filename)
```
6. For every variable in `glb.xml`, the `globalAnnual` is computed (the last 2 blocks have essentially just bypassed the function call of step 3):
```
git grep -n -B 3 "ts.globalAnnual" coupled_global.py
coupled_global.py-554-        ts = TS(exp[exp_key])
coupled_global.py-555-        for var in var_list:
coupled_global.py-556-            try:
coupled_global.py:557:                v, units = ts.globalAnnual(var)
--
coupled_global.py-792-        # Optionally read ohc
coupled_global.py-793-        if exp["ocean"] is not None:
coupled_global.py-794-            ts = TS(exp["ocean"])
coupled_global.py:795:            exp["annual"]["ohc"], _ = ts.globalAnnual("ohc")
--
coupled_global.py-798-
coupled_global.py-799-        if exp["vol"] is not None:
coupled_global.py-800-            ts = TS(exp["vol"])
coupled_global.py:801:            exp["annual"]["volume"], _ = ts.globalAnnual("volume")
```
7. `globalAnnual` is computed formerly by `cdutil.YEAR` and with this PR by `self.f.temporal.group_average`:
```
$ git grep -n group_average readTS.py
readTS.py:68:            v = self.f.temporal.group_average(v, "year")
```
</details>

> I assume this XML is used downstream with cdms2 (or another CDAT package) to open up the dataset.

Per the above, this is correct for `global_time_series`, but not for `e3sm_diags` (at least not directly. That is indeed probably the case within the `e3sm_diags` package -- but that is outside the scope of the `zppy` package).

> you can probably bypass the XML step entirely and just pass the glob paths

Ok thanks, I will try experimenting with this.

zppy/templates/readTS.py
```diff
@@ -62,10 +69,15 @@ def globalAnnual(self, var):
...
+            # Refactor note: `v = self.f(var)` gave `cdms2.tvariable.TransientVariable`
             units = v.units
 
             # Annual average
-            v = cdutil.YEAR(v)
+
+            # Refactor note: `AttributeError: 'Dataset' object has no attribute 'temporal'` seems to always occur
+            # Regardless if using CDAT or not, if using as object or class method.
+            # v = self.f.temporal.group_average(v, "year")
+            v = xarray.Dataset.temporal.group_average(v, "year")
```
forsyth2, line 81: @tomvothecoder It does indeed seem like I'm able to eliminate the `cdscan` call to generate an `xml` + the `self.f = cdms2.open(filename)` line by using `xarray.open_mfdataset(f"{directory}/*.nc")`

However, no matter what I do, Python doesn't seem to think `temporal` exists. I'm looking at https://xcdat.readthedocs.io/en/latest/generated/xarray.Dataset.temporal.group_average.html, and this is running in the latest Unified environment. Am I using the wrong averaging function here? (This line replaces `v = cdutil.YEAR(v)`)

forsyth2, line 81: @tomvothecoder It does indeed seem like I'm able to eliminate the `cdscan` call to generate an `xml` + the `self.f = cdms2.open(filename)` line by using `xarray.open_mfdataset(f"{directory}/*.nc")`

However, no matter what I do, Python doesn't seem to think `temporal` exists. I'm looking at https://xcdat.readthedocs.io/en/latest/generated/xarray.Dataset.temporal.group_average.html, and this is running in the latest Unified environment. Am I using the wrong averaging function here? (This line replaces `v = cdutil.YEAR(v)`)

tomvothecoder: Make sure to `import xcdat` to gain access to the `.temporal` accessor class from xCDAT.

The xCDAT accessor classes extend top of `xarray.Dataset` objects.

forsyth2: Oh I see, I just assumed it would be an `xarray` sub-module. Thanks for the info!

forsyth2:
@tomvothecoder After much debugging/learning of `xarray`'s types and data structures, I appear to actually have `global_time_series` working without any CDAT dependencies. 

`e3sm_diags` may still require the `cdscan` replacement to create `xml` file (unless the `e3sm_diags` package is updated similarly to take the glob path directly).

As a side note, it's a real shame we weren't using typing on _all_ `zppy` development, or all `e3sm_diags` development for that matter. Using explicit types really clears up if a variable is a CDAT remnant. In any case, using typing more going forward will certainly help clean up the code.

zppy/templates/coupled_global.py
```diff
@@ -3,12 +3,15 @@
...
-from typing import Any, List, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
+import cftime
 import matplotlib as mpl
 import matplotlib.backends.backend_pdf
 import matplotlib.pyplot as plt
 import numpy as np
+import xarray
+import xcdat  # noqa: F401
```
forsyth2, line 14: Not needed here

xylar:
> e3sm_diags may still require the cdscan replacement to create xml file (unless the e3sm_diags package is updated similarly to take the glob path directly).

I would hope that a similar approach would work in `e3sm_diags` and anywhere else using `cdscan`.  Is the issue only the difference between passing a file (e.g. the XML file produced by `cdscan`) that has a list of files vs. just passing the list of files?

tomvothecoder:
> `e3sm_diags` may still require the `cdscan` replacement to create `xml` file (unless the `e3sm_diags` package is updated similarly to take the glob path directly).

`e3sm_diags` doesn't actually use `cdscan` as a dependency. I validated this by performing a full code-base search and no results appeared (it's also not in the [dev.yml](https://github.com/E3SM-Project/e3sm_diags/blob/main/conda-env/dev.yml)).

What I _think_ `zppy` is doing is creating the XML linking _multi-file_ datasets (more than 1 `.nc` file) then passing it to `e3sm_diags` to be opened with `cdms2` internally. You probably have better knowledge though since this is the `zppy` codebase. Also pinging @chengzhuzhang to see if this sounds correct. 

If it is correct, then we need to refactor `e3sm_diags` to use Xarray/xCDAT first before refactoring `zppy` is possible. I think `cdms2` only supports opening multi-file datasets using `cdscan`, while Xarray can support the glob path or list of paths.

forsyth2:
>  Is the issue only the difference between passing a file (e.g. the XML file produced by `cdscan`) that has a list of files vs. just passing the list of files?

@xylar Yes, I believe that was the point of producing the `xml` files.

> What I _think_ `zppy` is doing is creating the XML linking _multi-file_ datasets (more than 1 `.nc` file) then passing it to `e3sm_diags` to be opened with `cdms2` internally.

@tomvothecoder Yes, sorry I should have been clearer. By "`e3sm_diags` may still require the `cdscan` replacement to create `xml` file ", I meant that _zppy_ would need the replacement in order to generate the xml file for `e3sm_diags`

> I think cdms2 only supports opening multi-file datasets using cdscan, while Xarray can support the glob path or list of paths.

@tomvothecoder Interesting, yes that's what I meant by "(unless the `e3sm_diags` package is updated similarly to take the glob path directly)." Ok, so it sounds like we want to go with that option then? That is, give `e3sm_diags` the ability to take in multiple files, rather than having zppy create a single file for it to use?

Basically the two options I was getting at were:
1. Create a `cdscan` replacement in zppy so zppy can pass a single xml file to e3sm_diags
2. Update e3sm_diags so it can just take a list of files (via glob path) directly

So it sounds like we want (2)

forsyth2:
> Installing a custom build of e3sm_diags with the cdat-migration-fy24 branch in a zppy dev env -- this is preferred since we can test earlier rather than later

@tomvothecoder Sure, that sounds good to me. Would it be better to merge the parts we can now though? I.e. `global_time_series` is now working without CDAT, so I would say we should just merge that now. And then we can do this test of `e3sm_diags` through `zppy` using the `cdat-migration-fy24`, but not actually merge that until `e3sm_diags` is in fact refactored on `main` (since otherwise it's not really useful and in fact breaks the current implementation)?

tomvothecoder:
> > Installing a custom build of e3sm_diags with the cdat-migration-fy24 branch in a zppy dev env -- this is preferred since we can test earlier rather than later
> 
> @tomvothecoder Sure, that sounds good to me. Would it be better to merge the parts we can now though? I.e. `global_time_series` is now working without CDAT, so I would say we should just merge that now. And then we can do this test of `e3sm_diags` through `zppy` using the `cdat-migration-fy24`, but not actually merge that until `e3sm_diags` is in fact refactored on `main` (since otherwise it's not really useful and in fact breaks the current implementation)?

Yes, I would split out the work to refactor `e3sm_diags` and `cdscan` in a separate branch since it will be long-standing. Work that can be merged into `main` can be done on their own branch(es) ahead of time.

PR #528, opened 2023-11-08
Bump to 2.3.1

forsyth2:
Bump to 2.3.1


forsyth2:
@xylar @tomvothecoder I just noticed I mistakenly bumped the version here to `2.3.1` rather than `2.3.1rc1`. Is that going to cause a problem for the actual release?

xylar:
@forsyth2, no worries.  It doesn't really matter what the version number is within the code for a release candidate.  What matters is that the version number is correct when you tag the release version.

forsyth2:
Ok awesome, thanks!

Issue #540, opened 2024-01-10
Fix test update commands


forsyth2:
### Request criteria

- [X] I searched the [zppy GitHub Discussions](https://github.com/E3SM-Project/zppy/discussions) to find a similar question and didn't find it.
- [X] I searched the [zppy documentation](https://e3sm-project.github.io/zppy).
- [X] This issue does not match the other templates (i.e., it is not a bug report, documentation request, feature request, or a question.)

### Issue description

In https://github.com/E3SM-Project/zppy/blob/main/tests/integration/template_directions.md, the line `rm -rf #expand user_www#forsyth/zppy_test_bundles_www/<UNIQUE ID>/v2.LR.historical_0201/image_check_failures` erroneously includes "forsyth". That should be removed.

forsyth2:
The same error appears for the complete-run test line.

forsyth2:
Additional changes:

```
# Remove the image check failures, so they don't end up in the expected files.
rm -rf /lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2/forsyth/zppy_test_complete_run_www/<UNIQUE ID>/v2.LR.historical_0201/image_check_failures
```
=>
```
# Remove the image check failures, so they don't end up in the expected files.
rm -rf image_check_failures
```

PR #547, opened 2024-02-06
Remove waves and eke from MPAS-Analysis generate option

xylar:
Typical E3SMv3 WCYCL and CRYO configurations will not have the output necessary to run waves and eddy kinetic energy (EKE) analysis.  It should be turned off to prevent confusing (though not fatal) errors in the setup stage of an MPAS-Analysis run.

### DIFF
--- a/zppy/templates/cryosphere.cfg
+++ b/zppy/templates/cryosphere.cfg
@@ -2,4 +2,4 @@
 sets= "lat_lon","zonal_mean_xy","zonal_mean_2d","polar","cosp_histogram","meridional_mean_2d","enso_diags","qbo","diurnal_cycle","annual_cycle_zonal_mean","streamflow","zonal_mean_2d_stratosphere","aerosol_aeronet","aerosol_budget"
 
 [mpas_analysis]
-generate = 'all', 'no_BGC', 'no_icebergs'
+generate = 'all', 'no_BGC', 'no_icebergs', 'no_waves', 'no_eke'
--- a/zppy/templates/default.ini
+++ b/zppy/templates/default.ini
@@ -257,7 +257,7 @@ anomalyRefYear = integer(default=1)
 cache = boolean(default=True)
 # Note that environment_commands needs to be the same for all related runs of [mpas_analysis].
 # For example, if years 1-50 are run using one environment and years 51-100 are run using another, MPAS-Analysis may fail.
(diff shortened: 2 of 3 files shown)


xylar:
@forsyth2, when you have a chance, could you suggest any testing I should do to verify that this fix actually works as expected?  It should be sufficient to do one run of watercycle and another of cryo analysis on Chrysalis (possibly just with MPAS-Analysis enabled).  Do you have any existing tests like that?

forsyth2:
@xylar Thanks for working on this.

- Note that `test_campaign.py` (https://github.com/E3SM-Project/zppy/blob/main/tests/integration/test_campaign.py) will definitely need updated values. (See https://github.com/E3SM-Project/zppy/blob/main/tests/integration/test_campaign.py)

- If you want to run the complete test suite, follow the "test dev (run before making a new zppy RC)" directions at https://e3sm-project.github.io/zppy/_build/html/main/dev_guide/release_testing.html. (I use those steps even when there is no pending RC; it's just useful for testing any PR).

- If you just want to see how the results look, you can run a smaller cfg. Perhaps something like the following:

```
[default]
campaign = "cryosphere" # TODO: Also test with "water_cycle"
case = v2.LR.historical_0201
constraint = ""
dry_run = "False"
environment_commands = ""
input = "/lcrc/group/e3sm/ac.forsyth2//E3SMv2/v2.LR.historical_0201"
input_subdir = archive/atm/hist
mapping_file = "map_ne30pg2_to_cmip6_180x360_aave.20200201.nc"
# TODO: Edit `output` and `www` to your own directories.
output = "/lcrc/group/e3sm/ac.forsyth2/zppy_test_complete_run_output/unique_id/v2.LR.historical_0201"
partition = "debug"
qos = "regular"
www = "/lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2/zppy_test_complete_run_www/unique_id"

[mpas_analysis]
active = True
anomalyRefYear = 1850
climo_years ="1850-1854", "1855-1860",
enso_years = "1850-1854", "1855-1860",
mesh = "EC30to60E2r2"
parallelTaskCount = 6
partition = "compute"
qos = "regular"
ts_years = "1850-1854", "1850-1860",
walltime = "00:30:00"
```

xylar:
@forsyth2, perfect!  That more simplified option is what I was after as a starting point.  The tip on updating `test_campaign.py` was also helpful!

xylar:
@forsyth2, yep, those tests are still running but they worked exactly as expected.  (Some tasks still fail during setup for the cryosphere configuration but that's because the run you use as a reference isn't really a cryosphere run, so some expected output is not available.)

Do you want me to run the tests and update the expected files once the PR has been approved?  Or do you prefer to do that yourself?

forsyth2:
> Do you want me to run the tests and update the expected files once the PR has been approved? Or do you prefer to do that yourself?

If it's not too much trouble, that sounds good, thanks! I usually consider a Chrysalis run good enough for PRs (I run the test suite on Perlmutter and Compy before RCs). Note that there are a couple things that need fixing in the testing directions (#540). There also might be some permissions issues that pop up. If it proves to be too much of a hassle, let me know and I can run the tests myself.

Issue #549, opened 2024-02-17
Address "flaky" `ts` subtask in `complete_run` test


forsyth2:
### Request criteria

- [X] I searched the [zppy GitHub Discussions](https://github.com/E3SM-Project/zppy/discussions) to find a similar question and didn't find it.
- [X] I searched the [zppy documentation](https://e3sm-project.github.io/zppy).
- [X] This issue does not match the other templates (i.e., it is not a bug report, documentation request, feature request, or a question.)

### Issue description

The `complete_run` integration test is "flaky" (e.g., as in https://docs.gitlab.com/ee/development/testing_guide/flaky_tests.html). Specifically, the `[[ ts ]] > [[ land_monthly ]]` subtask.

I was able to get the test passing while testing #424. However, it appears I never ran the test straight-through and instead did several re-runs, which actually had the effect of creating a _different_ final output than if it had been run straight through.

I came across this while testing #548. The subtask was suddenly failing despite no relevant changes being made. I then tested on the `main` branch (specifically at a9025491014b241449a14818a1c9bed98fdd8233)  and encountered the same errors. I found the following.

Run 1
------
Relevant sections from `tests/integration/generated/test_complete_run_chrysalis.cfg`:
```
[default]
case = v2.LR.historical_0201
constraint = ""
dry_run = "False"
environment_commands = ""
input = "/lcrc/group/e3sm/ac.forsyth2//E3SMv2/v2.LR.historical_0201"
input_subdir = archive/atm/hist
mapping_file = "map_ne30pg2_to_cmip6_180x360_aave.20200201.nc"
# To run this test, edit `output` and `www` in this file, along with `actual_images_dir` in test_complete_run.py
output = "/lcrc/group/e3sm/ac.forsyth2/zppy_test_complete_run_output/test-main2-20240216/v2.LR.historical_0201"
partition = "debug"
qos = "regular"
www = "/lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2/zppy_test_complete_run_www/test-main2-20240216"

[ts]
active = True
walltime = "00:30:00"
years = "1850:1854:2",

  [[ land_monthly ]]
  extra_vars = "landfrac"
  frequency = "monthly"
  input_files = "elm.h0"
  input_subdir = "archive/lnd/hist"
  vars = "FSH,RH2M"
  ts_fmt = "cmip"

[ilamb]
active = True
grid = '180x360_aave'
short_name = 'v2.LR.historical_0201'
[...]

PR #558, opened 2024-03-01
Specify year range for e3sm_to_cmip in `ts`

chengzhuzhang:
To address potential problem documented here: https://github.com/E3SM-Project/zppy/issues/549#issuecomment-1955728539

### DIFF
--- a/zppy/templates/ts.bash
+++ b/zppy/templates/ts.bash
@@ -141,7 +141,11 @@ cat > default_metadata.json << EOF
 EOF
 {
   export cmortables_dir={{ cmor_tables_prefix }}/cmip6-cmor-tables/Tables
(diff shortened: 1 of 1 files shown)


chengzhuzhang:
Hey @forsyth2, it is great news that the tests are all passing with the new e3sm_to_cmip release. However, I still think we should consider to use specified years for `e3sm_to_cmip` as discussed earlier. Do you think we should add this fix along the new rc?

forsyth2:
@chengzhuzhang Yes, I can implement that for the RC.

chengzhuzhang:
> @chengzhuzhang Oh sorry, I thought this was an issue, not a PR. I will test this fix too.

Thank you!

forsyth2:
All tests pass, merging.

Issue #552, opened 2024-02-21
Convert tests to use v3


forsyth2:
### Request criteria

- [X] I searched the [zppy GitHub Discussions](https://github.com/E3SM-Project/zppy/discussions) to find a similar question and didn't find it.
- [X] I searched the [zppy documentation](https://e3sm-project.github.io/zppy).
- [X] This issue does not match the other templates (i.e., it is not a bug report, documentation request, feature request, or a question.)

### Issue description

The tests currently use v2. Now that v3 simulation data is (or will soon be) available, we should convert the tests to use that instead. This will involve mapping file and grid changes, as in https://github.com/E3SM-Project/zppy/pull/548#issuecomment-1940414730.

forsyth2:
When this is implemented, we should reduce the test data size. The v2 data took up 24 terabytes, which is not an acceptable long-term solution. (See discussion on #561).

Issue #572, opened 2024-04-11
Improve zppy developer experience


forsyth2:
### Request criteria

- [X] I searched the [zppy GitHub Discussions](https://github.com/E3SM-Project/zppy/discussions) to find a similar question and didn't find it.
- [X] I searched the [zppy documentation](https://e3sm-project.github.io/zppy).
- [X] This issue does not match the other templates (i.e., it is not a bug report, documentation request, feature request, or a question.)

### Issue description

`zppy` has many complex inter-connected pieces. It can therefore be a challenging package to develop and debug (and debugging `zppy` could well mean debugging a package it calls).

There are some things we could do to improve the developer experience:

- [ ] Have `zppy` print out the `environment_commands` for each task on the `submitting...` line. This information is in the generated bash files, but this change would make this information hard to miss. This is important because a very common problem in `zppy` development is not having the environments for each task configured properly.
- [ ] Create a debugging guide for common challenges in debugging `zppy`
- [ ] Replacing the expected images tests with a metric checker test (#536). The latter is less prone to noise.
- [ ] Automated testing of the "combinatorial explosion" of parameter options. Often we need to know how `zppy` behaves when different combination of parameters are applied. Setting up automatic `cfg` generation and testing of these cases would be immensely helpful.
- [ ] More thorough testing in general, so more bugs are caught automatically.
- [ ] Nightly testing using `cron` or something like it. This will be necessary if the above points on more thorough testing are implemented, since `zppy` testing is already a lengthy process.

PR #608, opened 2024-06-19
Add minimal case cfgs

forsyth2:
Add minimal case cfgs. Addresses 'Automated testing of the "combinatorial explosion" of parameter options. Often we need to know how `zppy` behaves when different combination of parameters are applied. Setting up automatic `cfg` generation and testing of these cases would be immensely helpful.' in #572.

This pull request aims to add some important "minimal cases" -- i.e., cfgs with as few parameters as possible to test specific cases.

### DIFF
--- a/tests/integration/generated/directions_chrysalis.md
+++ b/tests/integration/generated/directions_chrysalis.md
@@ -145,3 +145,12 @@ cp -r /lcrc/group/e3sm/ac.forsyth2/zppy_test_bundles_output/<UNIQUE ID>/v2.LR.hi
 ```
 cp -r /lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2/zppy_test_complete_run_www/<UNIQUE ID>/v2.LR.historical_0201 /lcrc/group/e3sm/public_html/zppy_test_resources/expected_complete_run_unified_<#>
 ```
+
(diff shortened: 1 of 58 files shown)


forsyth2:
I've broken up the `complete_run` test into components or "minimal cases" as I call them here. 

I've also moved the input over from v2 to v3, in some sense resolving #552. To fully resolve #552, we'd need to have an actual test call (as #604 would do with the `complete_run` test) rather than just producing output (as these `cfg`s do).

My plan going forward is to use these minimal case `cfg`s to test future pull requests -- only running the relevant `cfg`s. E.g., no sense running ILAMB if I'm changing global_time_series code.

As for testing the `main` branch on some recurring basis, we need to either A) automate running every single one of these minimal cases and somehow checking their results, or B) convert the complete_run tests from v2 to v3 so we have _some_ method of testing `main`. (A) is more comprehensive, (B) is more practical.

Now, two caveats:

1. I haven't actually run these minimal cases because there are quite a few of them. My plan is to modify them as needed when actually using them to debug/test future pull requests.
2. The v3 input is currently hard coded to a simulation on Chrysalis. We'll need to copy over the input data to Perlmutter and Compy (or be content running these minimal cases on Chrysalis only).

PR #612, opened 2024-07-15
Remove climo_diurnal_input_files internal parameter

forsyth2:
Remove `climo_diurnal_input_files` internal parameter. Resolves #607

### DIFF
--- a/zppy/e3sm_diags.py
+++ b/zppy/e3sm_diags.py
@@ -126,16 +126,6 @@ def e3sm_diags(config, scriptDir, existing_bundles, job_ids_file):  # noqa: C901
             if "diurnal_cycle" in c["sets"]:
                 if c["dc_obs_climo"] == "":
                     c["dc_obs_climo"] = c["reference_data_path"]
-                diurnal_cycle_subtask_name = c["climo_diurnal_subsection"]
-                if diurnal_cycle_subtask_name in config["climo"].keys():
-                    if (
-                        "input_files"
-                        in config["climo"][diurnal_cycle_subtask_name].keys()
-                    ):
-                        diurnal_cycle_input_files = config["climo"][
-                            diurnal_cycle_subtask_name
-                        ]["input_files"]
-                        c["climo_diurnal_input_files"] = diurnal_cycle_input_files
             if ("streamflow" in c["sets"]) and (c["streamflow_obs_ts"] == ""):
                 c["streamflow_obs_ts"] = c["obs_ts"]
             print(prefix)
--- a/zppy/templates/default.ini
+++ b/zppy/templates/default.ini
@@ -127,7 +127,6 @@ backend = string(default="mpl")
 cfg = string(default="")
 # Name of the frequency from `[climo]` to use for "diurnal_cycle" runs
 climo_diurnal_frequency = string(default="")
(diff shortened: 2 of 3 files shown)


forsyth2:
Ran the following minimal cases (see #608):
```
e3sm_diags_diurnal_cycle
e3sm_diags_diurnal_cycle_mvm_1
e3sm_diags_diurnal_cycle_mvm_2
```
These produce results both with and without explicitly setting `{{ climo_diurnal_input_files }}` or `{{ climo_diurnal_input_files_ref }}`. 

@chengzhuzhang Is it better 1) to be explicit about the input files we're looking for (and thus require an extra parameter) or 2) to have zppy automatically determine the h number. (2) appears to work, but I'm concerned I just found a simple case where that happens to be true.

chengzhuzhang:
@forsyth2 thank you for working on this. To be explicit about the input files would be fine, I'm curious about the mechanism for zppy to automatically infer the h number?

forsyth2:
> I'm curious about the mechanism for zppy to automatically infer the h number?

The second commit currently in this PR: https://github.com/E3SM-Project/zppy/pull/612/commits/b116dcfd76618e3855e3cd4519ac0e256020e5b4
```
cp -s ${climo_diurnal_dir_source}/${nc_prefix}.{{ climo_diurnal_input_files_ref }}_*_${begin_year}??_${end_year}??_climo.nc .
cp -s ${climo_diurnal_dir_source}/${nc_prefix}.{{ climo_diurnal_input_files }}_*_${begin_year}??_${end_year}??_climo.nc .
```
could simply be replaced with
```
cp -s ${climo_diurnal_dir_source}/${nc_prefix}.*_*_${begin_year}??_${end_year}??_climo.nc .
```

But my concern is that the directories I'm working with just happen to have only the correct `h` number in them.

chengzhuzhang:
@forsyth2 I looked it up `e3sm_diags` code, I think as long as the the ref directory includes those seasonal mean diurnal cycle files, the h number doesn't matter. The file name just needs to start with the case name. We need to make sure the ref directory should include the correct symlinks.

forsyth2:
@chengzhuzhang I've updated the second commit (dcb9eb175ab51090cd1bf69febeb405224b084ca) and it looks like everything still works if we remove the input file specification in all cases for diurnal cycle. It looks like it was originally introduced in https://github.com/E3SM-Project/zppy/commit/aaaa61df13a41124c5b19ec3b85c011ccd2eb6c1#diff-ed02c5b26bf8ed8e97148b4f81cfdca9afc1706855f5a7aba9e0b7f3c47cee52. It seems the cleaner solution would have been to just use `*` in the first place rather than introducing a parameter.

forsyth2:
I've tested with:
- `zppy -c tests/integration/generated/test_min_case_e3sm_diags_diurnal_cycle_mvm_1_chrysalis.cfg` & ``zppy -c tests/integration/generated/test_min_case_e3sm_diags_diurnal_cycle_mvm_2_chrysalis.cfg`
- `zppy -c tests/integration/generated/test_min_case_e3sm_diags_diurnal_cycle_chrysalis.cfg` (model-vs-obs)

chengzhuzhang:
>It seems the cleaner solution would have been to just use `*` in the first place rather than introducing a parameter.

Yes, I think it is a better solution.

PR #654, opened 2024-12-18
Remove deprecated parameters

forsyth2:
## Issue resolution
- Closes #653. Associated with https://github.com/E3SM-Project/zppy-interfaces/pull/11.

Select one: This pull request is...
- [ ] a bug fix: increment the patch version
- [x] a small improvement: increment the minor version
- [ ] an incompatible (non-backwards compatible) API change: increment the major version

## 1. Does this do what we want it to do?

Objectives:
- Remove parameters that are no longer needed

Required:
- [x] Product Management: I have confirmed with the stakeholders that the objectives above are correct and complete.
- [ ] Testing: I have added or modified at least one "min-case" configuration file to test this change. Every objective above is represented in at least one cfg.
  - Weekly tests check if `zppy` works without these parameters.
- [ ] Testing: I have considered likely and/or severe edge cases and have included them in testing.
  - Weekly tests check if `zppy` works without these parameters.

## 2. Are the implementation details accurate & efficient?

Required:
- [x] Logic: I have visually inspected the entire pull request myself.
- [x] Logic: I have left GitHub comments highlighting important pieces of code logic. I have had these code blocks reviewed by at least one other team member.

## 3. Is this well documented?

Required:
- [x] Documentation: by looking at the docs, a new user could easily understand the functionality introduced by this pull request.
  - Parameters are documented in `default.ini`

## 4. Is this code clean?

Required:
- [x] Readability: The code is as simple as possible and well-commented, such that a new team member could understand what's happening.
- [x] Pre-commit checks: All the pre-commits checks have passed.

If applicable:
- [x] Software architecture: I have discussed relevant trade-offs in design decisions with at least one other team member. It is unlikely that this pull request will increase tech debt.

PR #657, opened 2025-01-14
Finish removing deprecated parameters

forsyth2:
## Issue resolution
- Follow-up to #654 (after running `git grep -n` on the 3 deprecated parameters)


forsyth2:
Note: `git grep -n` didn't show any remaining occurrences in `zppy-interfaces`, so no need for a corresponding PR there.

PR #661, opened 2025-01-17
Update conda recipe to be more like conda-forge feedstock

xylar:
## Issue resolution
- Closes #<ISSUE_NUMBER_HERE>

Please fill out either the "Small Change" or "Big Change" section, and delete the other.

## Small Change

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,5 +1,6 @@
 {% set name = "zppy" %}
 {% set version = "3.0.0rc1" %}
+{% set python_min = "3.9" %}
 
 package:
   name: {{ name|lower }}
@@ -11,17 +12,16 @@ source:
 
 build:
   number: 0
-  script: "{{ PYTHON }} -m pip install . --no-deps -vv"
+  script: {{ PYTHON }} -m pip install . --no-deps --no-build-isolation -vv
   noarch: python
-  entry_points:
-    - zppy = zppy.__main__:main
 
 requirements:
   host:
-    - python >=3.9
+    - python {{ python_min }}
     - pip
+    - setuptools
   run:
-    - python >=3.9
+    - python >={{ python_min }}
     - configobj >=5.0.0,<6.0.0
     - jinja2 >=2.0.0
     - mache >=1.3.2
@@ -29,6 +29,7 @@ requirements:
 
 test:
   requires:
+    - python {{ python_min }}
     - pip
   imports:
     - zppy


xylar:
## Testing

I did a test build of `zppy` on Chrysalis using the updated package.  It successfully built `zppy-3.0.0rc1-py_0.tar.bz2`.  I do not think further testing is necessary.

conda/meta.yaml
```diff
@@ -11,24 +12,24 @@ source:
 
 build:
   number: 0
-  script: "{{ PYTHON }} -m pip install . --no-deps -vv"
+  script: {{ PYTHON }} -m pip install . --no-deps --no-build-isolation -vv
```
xylar, line 15: It is best to not allow `pip` to install missing build dependencies from PyPI (which is what `--no-build-isolation` does).

The quotes are not needed.

conda/meta.yaml
```diff
@@ -11,24 +12,24 @@ source:
...
   noarch: python
-  entry_points:
-    - zppy = zppy.__main__:main
```
xylar, line 17: Entry points no longer need to be defined in conda recipes and removing them removes one more maintenance step we would otherwise have to do.

conda/meta.yaml
```diff
@@ -11,24 +12,24 @@ source:
...
 
 requirements:
   host:
-    - python >=3.9
+    - python {{ python_min }}
```
xylar, line 20: Here and below, conda-forge has moved to defining `python_min` (`3.9` by default) and ensuring that the package bets built and tested with this minimum version.  The idea is that thing tend to break more often when we only test with the newest python than when we only test with the oldest.  See https://github.com/conda-forge/cfep/blob/main/cfep-25.md for details.

conda/meta.yaml
```diff
@@ -11,24 +12,24 @@ source:
...
     - pip
+    - setuptools
```
xylar, line 22: `zppy` gets installed with `setuptools` so this is a necessary dependency.  It used to get bundled with `pip` but that is no longer the case.

forsyth2:
@xylar Thanks for working on this!

> It successfully built `zppy-3.0.0rc1-py_0.tar.bz2`. I do not think further testing is necessary.

Great, I see 3.0.0rc1 on https://anaconda.org/conda-forge/zppy/files. Was that created using this code, or do I need to merge this and make a `rc2`?

Also, perhaps I should have made the PR template clearer (#658), but my intention was if it was a small change, one could likely ignore most of the checkboxes (the numbered sections are sub-headings of "Big Change")

xylar:
> Great, I see 3.0.0rc1 on https://anaconda.org/conda-forge/zppy/files. Was that created using this code, or do I need to merge this and make a rc2?

No, neither.  This change is just for convenience and will not be tested anywhere.  I would suggest you merge it before we make rc2 (assuming we do) but no additional work or testing is required.

xylar:
> Also, perhaps I should have made the PR template clearer (https://github.com/E3SM-Project/zppy/pull/658), but my intention was if it was a small change, one could likely ignore most of the checkboxes (the numbered sections are sub-headings of "Big Change")

Oh, that definitely wasn't clear to me.  I would suggest adding a comment to that effect to the template.

PR #664, opened 2025-01-29
Update testing paths and commands

forsyth2:
## Summary

Objectives:
- Make sure all `environment_commands` variants get set to use Unified when doing unified testing
- Allow distinct parent directories for v2 and v3 testing data.

## Small Change


tests/integration/utils.py
```diff
@@ -280,7 +283,13 @@ def generate_cfgs(unified_testing=False, dry_run=False):
     )
     expansions = get_expansions()
     if unified_testing:
+        expansions["diags_environment_commands"] = expansions[
```
forsyth2, line 286: Might be useful to wrap this in a `if expansions["diags_environment_commands"] == "":`, so it doesn't overwrite an explicitly set `diags_environment_commands`. Currently testing with Unified only, so fine for now.

forsyth2: Addressed at https://github.com/E3SM-Project/zppy/pull/676

PR #673, opened 2025-02-04
Deprecated parameter alerts

forsyth2:
## Summary

Objectives:
- Error if the user is trying to use a deprecated parameter.

Issue resolution:
- Addresses https://github.com/E3SM-Project/zppy/discussions/667#discussioncomment-12047688

## Small Change

### DIFF
--- a/zppy/utils.py
+++ b/zppy/utils.py
@@ -31,6 +31,10 @@ class DependencySkipError(RuntimeError):
     pass
 
 
+class DeprecatedParameterError(RuntimeError):
+    pass
+
+
 # Utitlities for this file ####################################################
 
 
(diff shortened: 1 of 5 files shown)


zppy/utils.py
```diff
@@ -323,6 +327,17 @@ def check_parameter_defined(c: Dict[str, Any], relevant_parameter: str) -> None:
...
 
+def check_for_deprecated_parameters(
+    c: Dict[str, Any], deprecated_parameters: List[str]
+) -> None:
+    present_parameters = []
+    for parameter in deprecated_parameters:
+        if parameter in c.keys():
+            present_parameters.append(parameter)
+    if present_parameters:
+        raise DeprecatedParameterError(present_parameters)
```
forsyth2, line 338: We could also just do a warning, but I figured an error is more likely to get the user's attention.

forsyth2 (commented):
@chengzhuzhang This is ready for review.

zppy/utils.py
chengzhuzhang: does  `ts_fmt` a no-op or raise error before this change?

forsyth2: on `main`, it just won't get used by anything. That is, if a user sets it thinking it's useful, they'll get no indication (either warning or error) that it's not actually useful. (Confirmed with cfg in https://github.com/E3SM-Project/zppy/pull/677).

forsyth2:
@chengzhuzhang Did we decide we want these changes? If so, I'll go ahed and merge.

Issue #669, opened 2025-02-03
[Bug]: Another parameter checking error: zppy.utils.ParameterNotProvidedError: climo_land_subsection


chengzhuzhang:
### What happened?

I think it is also something introduced in v3. My config is as follows:
```
[e3sm_diags]
active = True
walltime = "4:00:00"
#years = "0001:0100:50", "0001:0100:100"
years = "1985:2014:30",
ts_num_years = 30
ref_start_yr = 1985
ref_final_yr = 2014
multiprocessing = True
num_workers = 8

  [[ lnd_monthly_mvm_lnd ]]
  # Test model-vs-model using the same files as the reference
  #environment_commands = "source /home/ac.zhang40/y/etc/profile.d/conda.sh; conda activate edv290"
  grid = 'native'
  climo_subsection = "land_monthly_climo"
  diff_title = "Difference"
  partition = "compute"
  qos = "regular"
  short_name = v3.LR.piControl
  ref_name = "20231209.v3.LR.piControl-spinup.chrysalis"
  ref_start_yr = 0051
  ref_final_yr = 0100
  ref_years = "0051-0100",
  reference_data_path = "/lcrc/group/e3sm/ac.zhang40/tests/20231209.v3.LR.piControl-spinup.chrysalis_land_diags/post/lnd/native/clim"
  run_type = "model_vs_model"
  sets = "lat_lon_land",
  short_ref_name = "20231209.v3.LR.piControl-spinup"
  swap_test_ref = False
  tag = "model_vs_model"
  ts_num_years_ref = 50
```

```
URL: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.zhang40/E3SMv3_eurc5/v3.LR.historical_0051/e3sm_diags
e3sm_diags_lnd_monthly_mvm_lnd_model_vs_model_1985-2014_vs_0051-0100
Traceback (most recent call last):
  File "/home/ac.zhang40/.conda/envs/zppyv3/bin/zppy", line 8, in <module>
    sys.exit(main())
  File "/home/ac.zhang40/.conda/envs/zppyv3/lib/python3.9/site-packages/zppy/__main__.py", line 61, in main
    _launch_scripts(config, script_dir, job_ids_file, plugins)
  File "/home/ac.zhang40/.conda/envs/zppyv3/lib/python3.9/site-packages/zppy/__main__.py", line 237, in _launch_scripts
[...]

PR #675, opened 2025-02-04
Explain parameter errors

forsyth2:
## Summary

Objectives:
- Explain parameter errors

Issue resolution:
- Addresses https://github.com/E3SM-Project/zppy/issues/669#issuecomment-2632486844

## Small Change


forsyth2 (commented):
@chengzhuzhang Would these changes help with understanding the sources of parameter errors, as in https://github.com/E3SM-Project/zppy/issues/669?

That error, for example, would now give "climo_land_subsection is needed because the set `lat_lon_land` requires it".

PR #659, opened 2025-01-16
Skip `oceanConservation` task in MPAS-Analysis

xylar:
The conservation check is turned off in E3SM by default so we don't want to run the analysis by default.  (Doing so would ultimately be harmless but there would be a confusing error message on setup.)

## Issue resolution

Select one: This pull request is...
- [ ] a bug fix: increment the patch version
- [x] a small improvement: increment the minor version
- [ ] a new feature: increment the minor version
- [ ] an incompatible (non-backwards compatible) API change: increment the major version

## Small Change

- [x] To merge, I will use "Squash and merge". That is, this change should be a single commit.

### 1. Does this do what we want it to do?

Objectives:
- Prevent confusing error messages from MPAS-Analysis when the conservation check is not enabled in E3SM runs

Required:
- [x] Product Management: I have confirmed with the stakeholders that the objectives above are correct and complete.
- [ ] Testing: I have added or modified at least one "min-case" configuration file to test this change. Every objective above is represented in at least one cfg.
- [ ] Testing: I have considered likely and/or severe edge cases and have included them in testing.

### 2. Are the implementation details accurate & efficient?

Required:
- [x] Logic: I have visually inspected the entire pull request myself.
- [ ] Logic: I have left GitHub comments highlighting important pieces of code logic. I have had these code blocks reviewed by at least one other team member.

### 3. Is this well documented?

Required:
- [x] Documentation: by looking at the docs, a new user could easily understand the functionality introduced by this pull request.

### 4. Is this code clean?

Required:
- [x] Readability: The code is as simple as possible and well-commented, such that a new team member could understand what's happening.
- [ ] Pre-commit checks: All the pre-commits checks have passed.

PR #678, opened 2025-02-05
Remove MPAS-Analysis ocean conservation from default config

xylar:
We do not plan to run with the conservation check on by default, so it doesn't make sense to have the analysis plots on by default.

Additionally, the v3 reference simulation is outputting conservation at a monthly interval, but MPAS-Analysis assumes it is being output at a daily interval, the new default.  Incorrect rolling averages were leading to no data being plotted.

## Summary

Objectives:
- Fix MPAS-Analysis plots of ocean conservation in E3SM-Unified 1.11.0rc5 that are empty (see https://github.com/E3SM-Project/zppy/discussions/634#discussioncomment-12020270)
- Disable ocean conservation plots by default because the required data will not be available in most E3SM v3 simulations.

- [x] a bug fix: increment the patch version
- [ ] a small improvement: increment the minor version

## Small Change

### DIFF
--- a/zppy/defaults/default.ini
+++ b/zppy/defaults/default.ini
@@ -276,7 +276,7 @@ anomalyRefYear = integer(default=1)
 cache = boolean(default=True)
 # Note that environment_commands needs to be the same for all related runs of [mpas_analysis].
 # For example, if years 1-50 are run using one environment and years 51-100 are run using another, MPAS-Analysis may fail.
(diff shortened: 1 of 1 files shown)


xylar:
@forsyth2, please let me know if this change makes sense to you.  I think it should have been included in #659 and it was just an oversight.

forsyth2:
Yeah, this makes sense, thanks!

forsyth2:
I'll go ahead and merge this, thanks!

PR #733, opened 2025-09-04
Enable test customization

forsyth2:
## Summary

Objectives:
- Enable test customization by allowing the tester to choose which `cfg`s to run and which sets to run the image checker on. This will be quite useful when one test fails and we want to re-run it without running all the other jobs and/or tests unnecessarily. 

Issue resolution:
- N/A

## Small Change


tests/integration/utils.py
```diff
@@ -1,81 +1,30 @@
...
+    "global_time_series_environment_commands": "source <INSERT PATH TO CONDA>/conda.sh; conda activate <INSERT ENV NAME>",
+    "tests_to_run": [
+        "weekly_bundles",
+        "weekly_comprehensive_v2",
+        "weekly_comprehensive_v3",
+        "weekly_legacy_3.0.0_bundles",
+        "weekly_legacy_3.0.0_comprehensive_v2",
+        "weekly_legacy_3.0.0_comprehensive_v3",
+    ],
+    "sets_to_run": ["e3sm_diags", "mpas_analysis", "global_time_series", "ilamb"],
```
forsyth2, line 25: I guess this might be better labeled "tasks_to_run"

tests/integration/utils.py
```diff
@@ -1,81 +1,30 @@
...
-# Wait
-
-ls tests/integration/test_*.py # to see what tests are available to run
-pytest tests/integration/test_images.py
-# 1 passed in 2910.93s (0:48:30)
-"""
+TEST_SPECIFICS: Dict[str, Any] = {
+    "diags_environment_commands": "source <INSERT PATH TO CONDA>/conda.sh; conda activate <INSERT ENV NAME>",
+    "global_time_series_environment_commands": "source <INSERT PATH TO CONDA>/conda.sh; conda activate <INSERT ENV NAME>",
+    "tests_to_run": [
```
forsyth2, line 17: I guess this might be better labeled "cfgs_to_run"

forsyth2 (commented):
@chengzhuzhang Let me know if you want to review this test-side-only change. Otherwise, I think this should be good to merge (aside from a couple `dict` keys I think I'll rename first).

Key points:
- Most of the things that need to be changed between test runs are now at the very top of the `utils` file. That is, the environments, the `cfg`s to run, the tasks/sets to run, and the unique test run id.
- This change allows a tester to auto-generate _only_ the requested cfgs and to run the image checker on _only_ the requested results directories & task subdirectories. This is currently possible via cumbersome commenting of the test code. These changes make it easier to toggle that functionality on/off. This would be useful for example when most of the tests are passing, but we just want to test a small change in `ilamb` (i.e., no need to re-run `e3sm_diags` too).
- I tested this commit on `mpas_analysis`, `global_time_series`, and `ilamb` during my testing of https://github.com/E3SM-Project/zppy-interfaces/pull/26. I only left out `e3sm_diags` because of the diffs noted in the thread at https://github.com/E3SM-Project/zppy/discussions/634#discussioncomment-14194199. As expected, `e3sm_diags` tests were skipped rather than failed.

forsyth2:
> Let me know if you want to review this test-side-only change. Otherwise, I think this should be good to merge 

@chengzhuzhang I just want to re-ping you on this. I'd like to include this change for this week's weekly testing and other PR testing.

chengzhuzhang (approved):
I had a visual check, the changes look good for me.

PR #751, opened 2025-10-21
Update test_images

forsyth2:
## Summary

Objectives:
- Add some improvements to `test_images.py`

Issue resolution:
- Follow-up to https://github.com/E3SM-Project/zppy/discussions/746#discussioncomment-14735060

Select one: This pull request is...
- [ ] a bug fix: increment the patch version
- [x] a small improvement: increment the minor version
- [ ] a new feature: increment the minor version
- [ ] an incompatible (non-backwards compatible) API change: increment the major version

## Small Change

- [x] To merge, I will use "Squash and merge". That is, this change should be a single commit.
- [x] Logic: I have visually inspected the entire pull request myself.
- [x] Pre-commit checks: All the pre-commits checks have passed.

PR #752, opened 2025-10-21
Multithread image tests

### LINKED ISSUE #724: [Feature]: Parallelize image checker

The image checker test can take ~30 minutes to run. This isn't the major bottleneck when the prerequisite  `zppy` runs take over an hour, but it is still a substantial chunk of time. Furthermore, this time will increase as we add more `cfg`s to the Weekly testing suite. E.g., #720 adds 3 legacy `cfg`s.

The image checker test could be sped up by running each test in parallel. Possible implementations:
1. Make a separate `.py` for each `cfg` and run them individually with something like `srun -N1 pytest my_file.cfg`
2. Use Python's `Process` or similar class to launch multiple jobs in parallel.

_No response_

### Additional context

_No response_

forsyth2:
## Summary

Objectives:
- Multithread the image tests to reduce runtime.

Issue resolution:
- Closes #724

## Small Change

### DIFF
--- a/tests/integration/test_images.py
+++ b/tests/integration/test_images.py
@@ -232,7 +232,7 @@ def test_images():
         print(f"Running {len(test_configs)} tests in parallel")
         print("Individual test logs will be written to test_<name>.log files")
         with ProcessPoolExecutor(
-            max_workers=6, mp_context=get_context("spawn")
+            max_workers=6, mp_context=get_context("forkserver")
         ) as executor:
             # Submit all tests
             future_to_test = {


forsyth2:
Rebased after merging #751

forsyth2 (commented):
The code now appears to be working.

I was able to run with `1 passed in 946.73s (0:15:46)` rather than the previous `1 passed in 3566.36s (0:59:26)`. So, that's roughly a 4x speedup.

Before merging this though, we should:
- [x] Reset `test/integration/utils.py`
- [ ] Do a thorough visual code inspection. Claude was very helpful in drafting this, but we should do a manual inspection of the code it produced.
- [ ] Determine how to best run this (e.g., from a login node or by allocating a compute node)

tests/integration/test_images.py
```diff
@@ -17,128 +20,268 @@ def intersect_tasks(
...
+                expansions,
+                diff_dir_suffix,
+                tasks_to_run,
                 test_results_dict,
             )
+
+        output = captured_output.getvalue()
+
+        # Write to individual log file
+        log_filename = f"test_{test_name}.log"
```
forsyth2: Maybe make this `logs/...` so the logs are in a subdir for easy deletion (rather than `rm test_*.log`) or review.

forsyth2 (commented):
Visual inspection looks mostly ok; made a couple notes.

I'm thinking we can test this alongside the known working tests during this E3SM Unified testing period. If it doesn't encounter any further problems, it should be good to merge. Since it's just test code and nothing user-facing, it doesn't particularly matter if it's included in the final release or not.

PR #753, opened 2025-10-24
Fixes for rc3

forsyth2:
## Summary

Objectives:
- Ensure tests pass on E3SM Unified 1.12.0rc3

Issue resolution:
- Follow-up to https://github.com/E3SM-Project/zppy/pull/748

- [x] a bug fix: increment the patch version
- [ ] a small improvement: increment the minor version

## Small Change

### DIFF
--- a/zppy/templates/ts.bash
+++ b/zppy/templates/ts.bash
@@ -64,7 +64,7 @@ if grep -q "*" input.txt; then
 fi
 # Generate time series files
 # If the user-defined parameter "vars" is "", then ${vars}, defined above, will be too.
-cat input.txt | /home/ac.zender/bin_chrysalis/ncclimo --npo \
+cat input.txt | ncclimo \
 -c {{ case }} \
 {%- if vars != '' %}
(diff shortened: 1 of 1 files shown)


forsyth2:
> Copy/transfer the Chrysalis expected results to Perlmutter & Compy and test against those. This is probably the more robust approach, unless the plotting packages produce results that differ greatly depending between machines.

I think this would actually be infeasible.

```bash
# cd into the expected results directory on Chrysalis
cd /lcrc/group/e3sm/public_html/zppy_test_resources

# Let's count how many times `lcrc` appears
grep -R -l "lcrc" * | wc -l
9465

# And how many times a `ac.<username>` path appears
grep -R -l "ac\." * | wc -l
# Still going
```

In theory, we could systematically change the file paths per machine, but that's outside the scope of the Unified testing period. Therefore, we'll continue with what we've been doing, i.e., option 1:
> Examine the diffs, checking if it's likely all of them are due to changes over this development period. If so, update the local machine's expected results.

PR #760, opened 2025-11-25
Add additional files to the mpas-analysis cache for time series analysis

cbegeman:
## Summary

Objectives:
- Cache additional MPAS-Analysis files needed to extend time series plots from existing analysis when the source files are not present in the output path (typically because they have been long-term archived)
- Remove an MPAS-Analysis file that may be large (with `nCells` dimensions) when MPAS-Analysis has been completed which is not needed to extend time series plots

- [X] a small improvement: increment the minor version

## Small Change

- [X] To merge, I will use "Squash and merge". That is, this change should be a single commit.
- [X] Logic: I have visually inspected the entire pull request myself.
- [X] Pre-commit checks: All the pre-commits checks have passed.

### DIFF
--- a/zppy/templates/mpas_analysis.bash
+++ b/zppy/templates/mpas_analysis.bash
@@ -40,11 +40,6 @@ mkdir -p ${identifier}
 mkdir -p cfg
 
 {% if cache == true %}
-files=( "mpasIndexOcean.nc" "mpasTimeSeriesOcean.nc" "seaIceAreaVolNH.nc" "seaIceAreaVolSH.nc")
-for file in "${files[@]}"
-do
-  cp cache/timeseries/${file} ${identifier}/timeseries/${file}
-done
 # Restore cached copies of pre-computed files
 cached=( "timeseries/moc" "timeseries/OceanBasins" "timeseries/transport" )
 mkdir -p cache
@@ -52,6 +47,11 @@ for subdir in "${cached[@]}"
 do
   mkdir -p cache/${subdir} ${identifier}/${subdir}
   rsync -av cache/${subdir}/ ${identifier}/${subdir}/
+done
+files=( "mpasIndexOcean.nc" "mpasTimeSeriesOcean.nc" "seaIceAreaVolNH.nc" "seaIceAreaVolSH.nc")
+for file in "${files[@]}"
+do
+  cp cache/timeseries/${file} ${identifier}/timeseries/${file}
 done
 {% endif %}
 


xylar:
Okay, the environment and activation script are in place:
```
source /lcrc/soft/climate/e3sm-unified/load_hr_mpas_analysis.sh
```

forsyth2:
Note: I'll be merging #763 to update the test suite accordingly (tests needed to be updated to use a custom environment for `mpas_analysis`).

cbegeman:
@forsyth2 Thanks, I suspected they were unrelated. The pickle error may need to be resolved in the MPAS-Analysis environment but @xylar would know better.

xylar:
> Then, what would be doing the pickling?

It turns out that `multiprocessing` is doing the pickling, though it isn't clear to me what is causing an error, what might have changes, and why my own testing didn't reveal this.

forsyth2:
> It looks like you all need to be setting `mapMpiTasks = 1` for these runs.

Should that be true in general for standard `zppy` testing, or specifically for when we're testing a dev environment?

xylar:
> Should that be true in general for standard zppy testing, or specifically for when we're testing a dev environment?

This should only be true when ESMF is installed from conda, so only for these specialized, non-Unified MPAS-Analysis environments.

cbegeman:
@forsyth2 Ditto, thank you for testing! I agree with you that these changes are not significant.
@xylar Thank you for testing and creating the new mpas-a environment.

I'm not a zppy maintainer. @forsyth2 can you assign this to yourself or someone else?

forsyth2:
@chengzhuzhang re: [your earlier comment](https://github.com/E3SM-Project/zppy/pull/760#issuecomment-3603515715), would you like to do further testing on this before I merge?

chengzhuzhang:
@forsyth2 I was in contact with @wlin7 to see if we need a real test case to confirm this PR addresses the v3.HR use case.

chengzhuzhang (approved):
@wlin7 and @xylar, with the new commit, running mpas-analysis through zppy created reasonable time-series. Please review:
1.The testing setup: https://github.com/E3SM-Project/zppy/pull/760#issuecomment-3635013433
2.The results [here](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.zhang40/tests/E3SMv3_testings_mpas_cache_try7/v3.LR.historical_0051/mpas_analysis/ts_1850-1859_climo_1855-1859/)
3.The output directory: /lcrc/group/e3sm2/ac.zhang40/E3SMv3_testings/v3.LR.historical_0051_try7

Please review if this PR provided expected behavior, also note implication on other dependency. One thing I checked is that the global-time-series plot would use files from cache/timeseries, as long as this directory structure remains, the downstream plotting for global-time-series won't be impacted.

However, there is one important item to note, to go with deletion for mpas history file. 
Among the three global mean time-series plots: tracks changes of ocean heat content, sea-level, moc: the moc plot uses file from mpas-analysis cache, however the other two are computed from mpas history files (code see [here](https://github.com/E3SM-Project/zppy-interfaces/blob/bb080aa79f10c8a7b59450063457949eb2c34b38/zppy_interfaces/global_time_series/create_ocean_ts.py#L130))

zppy/templates/mpas_analysis.bash
```diff
@@ -297,10 +302,16 @@ fi
...
+for file in "${files[@]}"
+do
+  cp ${identifier}/timeseries/${file} cache/timeseries/${file}
+done
 for subdir in "${cached[@]}"
 do
   rsync -av ${identifier}/${subdir}/ cache/${subdir}/
 done
+# Remove one particularly large file which does not need to be cached
+rm ${identifier}/timeseries/mpasTimeSeriesSeaIce.nc
```
xylar: This seems to just be deleting a time-series file form the analysis directory, and not to be related to caching at all.  Is this perhaps a relic of an earlier approach to copying cache files and something no longer needed?  I would think it wouldn't be a great idea to delete output from the mpas-analysis output directory even if it is large unless we decide on a systematic way of cleaning things up that doesn't lose important data.

xylar (changes requested):
One line I think probably should be removed.

@chengzhuzhang, can your comments be resolved at this point?

xylar:
> However, there is one important item to note, to go with deletion for mpas history file.
Among the three global mean time-series plots: tracks changes of ocean heat content, sea-level, moc: the moc plot uses file from mpas-analysis cache, however the other two are computed from mpas history files (code see [here](https://github.com/E3SM-Project/zppy-interfaces/blob/bb080aa79f10c8a7b59450063457949eb2c34b38/zppy_interfaces/global_time_series/create_ocean_ts.py#L130))

@chengzhuzhang, I didn't quite follow what the solution is to this problem.  Do you think the MPAS-Ocean history files for ocean heat content and sea level need to be retained in order for global time series analysis to succeed?  Or is the suggestion to skip those variables? Or something else? It seems like it will be a problem if history files have to be retained for those plots, and also if those plots have to be skipped.

zppy/templates/mpas_analysis.bash
cbegeman: Removing large files from MPAS-Analysis that are not needed was requested by HR (specifically @wlin7). Do you think that MPAS-Analysis would be a better place to do this clean-up? I'm not particularly tied to including this in this PR but maybe some more context would be helpful:

This file seems to be somewhat of an outlier in MPAS-Analysis output files in that it has dimensions of Time (equal the length of the requested time series) and nCells (equal to the size of the whole mesh, which seems particularly egregious given that only a portion of those cells would contain sea ice). Correct me if there are other files like this. It's not clear to me that this would be "losing important data" because it's just copying information from history files that would (ideally) be archived as opposed to derived analysis variables.

```
ncinfo timeseries/mpasTimeSeriesSeaIce.nc
...
    dimensions(sizes): Time(72), StrLen(64), nCells(465044)
    variables(dimensions): |S1 xtime_startMonthly(Time, StrLen), |S1 xtime_endMonthly(Time, StrLen), float32 timeMonthly_avg_iceAreaCell(Time, nCells), float32 timeMonthly_avg_iceVolumeCell(Time, nCells), float32 timeMonthly_avg_snowVolumeCell(Time, nCells)

Comment (xylar) [reply zppy/templates/mpas_analysis.bash]: Yes, there is a corresponding file for the the ocean as far as I know.

**Comment by cbegeman** [comment]
@wlin7 and @chengzhuzhang Thank you both for your review and testing! I think this is good to merge when @wlin7 approves.

**Comment by chengzhuzhang** [comment]
Thank you @wlin7 for reviewing and confirming!

**Comment by forsyth2** [comment]
Thanks @cbegeman for working on this. Thanks @chengzhuzhang, @xylar, @wlin7 for reviewing/testing. I also just merged the associated test update (#763)

**Comment by forsyth2** [comment]
@xylar For `zppy`'s weekly `main` branch testing, now that this PR & https://github.com/MPAS-Dev/MPAS-Analysis/pull/1126 have been merged, should I be setting
```
environment_commands=source /lcrc/soft/climate/e3sm-unified/load_hr_mpas_analysis.sh
```
for the `mpas_analysis` task? (That is, the Unified `mpas_analysis` would no longer work?)

xylar:
@forsyth2 , I don't think so. I think the HR branch is unofficial and shouldn't become the new default for testing.

xylar:
@forsyth2 , I don't think so. I think the HR branch is unofficial and shouldn't become the new default for testing.

PR #763, opened 2025-12-02
Allow tests to use different `mpas_analysis` environment

forsyth2:
## Summary

Objectives:
- Add the ability to specify `environment_commands` for `mpas_analysis` in testing. (Note this was always possible as a user: just set `environment_commands` in the cfg).

Issue resolution:
- Allows for testing of #760.

## Small Change

### DIFF
--- a/tests/integration/generated/test_weekly_comprehensive_v2_chrysalis.cfg
+++ b/tests/integration/generated/test_weekly_comprehensive_v2_chrysalis.cfg
@@ -170,6 +170,7 @@ active = True
 anomalyRefYear = 1980
 climo_years ="1980-1984", "1985-1990",
 enso_years = "1980-1984", "1985-1990",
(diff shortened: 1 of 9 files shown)


forsyth2 (commented):
@chengzhuzhang I think this should be fine to merge.

A couple notes:
- The bundle cfgs weren't updated because they don't have a `mpas_analysis` task set up.
- I did update the legacy cfgs, which we normally think of as "frozen", but the change is the _value_ of `environment_commands`, not the parameters themselves. So, it's really no different than, for example, running the latest `e3sm_diags` dev environment on a legacy cfg.

chengzhuzhang:
okay, I thought specifying `environment_commands` was already enabled. This PR is just for testing. If it is the case, perhaps to update the title to avoid confusion.

forsyth2:
These are sort-of meta-environment commands -- i.e., `<task_name>_environment_commands` is really saying "what should we set as `environment_commands` in each of the auto-generated test cfgs for task `<task_name>`?"

PR #727, opened 2025-07-08
Applying computer vision to image checking

forsyth2:
# Hackathon project: Applying computer vision to image checking

This is a prototype and is not meant to be merged. The pull request is for easy visibility/ability to comment.

## Goal

The image checker can produce a great number of diffs. These are currenty organized by plot type. 

This poses a problem if many of the diffs are similar -- it's not a great use of a developer's time to manually look through say 1,000 diffs only to realize every single one is just "the pixels shifted slightly left" or "the colorbar changed so of course the plots look different".

What would be ideal is to use computer vision/machine learning to group those diffs how we want them. E.g., "every diff in `cluster_0/` is because the pixels shifted, every diff in `cluster_1/` is because its colorbar changed, and so on".

That would mean the developer only needs to look at a diff or two in each cluster to know what the issues are. There would be no need to manually review each diff, saving a lot of time.

## Setting up

<details>

<summary> setup </summary>

Set up the directories to test on:
```bash
# Set up actual_images_dir
cd /lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2
mkdir -p cv_prototype/actual_from_20250613/e3sm_diags/
cp -r zppy_weekly_comprehensive_v3_www/test_weekly_20250613/v3.LR.historical_0051/e3sm_diags/atm_monthly_180x360_aave cv_prototype/actual_from_20250613/e3sm_diags/
ls cv_prototype/actual_from_20250613/e3sm_diags/atm_monthly_180x360_aave/model_vs_obs_1987-1988/
# Contains the different sets, good

# Set up expected_images_dir
cd /lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2
mkdir -p cv_prototype/expected_from_unified/e3sm_diags/
cp -r /lcrc/group/e3sm/public_html/zppy_test_resources_previous/expected_results_for_unified_1.11.1/expected_comprehensive_v3/e3sm_diags/atm_monthly_180x360_aave cv_prototype/expected_from_unified/e3sm_diags/
ls cv_prototype/expected_from_unified/e3sm_diags/atm_monthly_180x360_aave/model_vs_obs_1987-1988/
# Contains the different sets, good

# Get the list of files we expect, expected_images_list
cd cv_prototype/expected_from_unified/
find . -type f -name '*.png' > ../image_list_expected.txt
cd ..
ls
# actual_from_20250613  expected_from_unified  image_list_expected.txt
```

To run:
```bash
# Initial setup
[...]

PR #773, opened 2026-01-21
Datathon Project: config builder

forsyth2:
This PR should NOT be merged. It is to show a prototype developed for the [January 2026 LLNL Datathon](https://data-science.llnl.gov/latest/llnl-datathon). See [here](https://github.com/E3SM-Project/zppy/pull/727) for an earlier (unrelated) zppy "hackathon" project.

Goal: a user should be able to give a simulation data directory path and get back a zppy cfg that if not perfect, at least gives a really good starting point. 

Motivating need: the zppy cfg is quite complex, with a very large number of parameters. It would therefore be useful to create custom starter cfg files for users to build on.

Architecture: This implementation consists of 3 layers, building on top of each other.
1. A script to extract available data from a simulation output directory. That's `simulation_output_reviewer.py`.
2. A script to generate a starting point cfg file based on the data found. That's `zppy_config_generator.py`.
3. An agentic AI question-answering system that would allow users to customize that cfg (the "answer" would be a better cfg than the starting point). Examples: a user could request "diagnostics relating to this particular physical phenomena" or "diagnostics on a hindcast during these particular years"

Layer 3 is more exploratory and the core of the Datathon challenege. Layers 1-2 will likely be cleaned up and merged into `zppy` as a distinct PR at a later date.

I've gone with this approach because a lot of the zppy cfg can in fact be constructed in a deterministic manner, lending itself more to a script than an AI agent. However, the more natural-language oriented layer 3 is a better fit for an AI agent, and the agent would also have a more-informed starting point rather than needing to "learn" all the rules of our data structuring.


config_builder/run_agent.py
```diff
@@ -0,0 +1,364 @@
...
+    args = parser.parse_args()
+
+    # Verify ollama is available
+    try:
+        subprocess.run(["ollama", "list"], capture_output=True, check=True)
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        print("Error: ollama not found. Please install ollama first:", file=sys.stderr)
+        print("  https://ollama.ai", file=sys.stderr)
+        print(
+            "  Official instructions: https://ollama.com/download. For Linux, that shows `curl -fsSL https://ollama.com/install.sh | sh`. After installation, verify that Ollama is available by running `ollama --version`",
```
forsyth2: Following directions from @tomvothecoder in https://github.com/aims-group/llnl-datathon-2026/pull/1/files: 

```
### 3. Option B: Local LLM via Ollama (Optional)

Ollama is used as a **local LLM service** for agentic workflows.

> **Important:** Ollama is **not managed by Conda**.
> It must be installed at the system level.
Install Ollama by following the official instructions:
[https://ollama.com/download](https://ollama.com/download)
```

forsyth2 (commented):
The [third commit](https://github.com/E3SM-Project/zppy/pull/773/commits/08fbe9910fc99391fde0a2999c31846075480f52) is an initial implementation of an agentic AI workflow. `run_agent.py` was largely generated with Claude, with fixes added by me. I will now proceed with actually installing `ollama` and testing it out.

## Notes from Claude on `Ollama`:

Usage Limits - Completely Unlimited! 🎉
Ollama is 100% free and open-source with NO limits:

✅ No token limits - Process as much as you want
✅ No API costs - Runs locally on your machine
✅ No internet required - After downloading models, works offline
✅ No rate limits - Run as many queries as you need
✅ Privacy - Your data never leaves your machine

Requirements:

Disk space: ~4.7GB for llama3.1:8b, ~40GB for llama3.1:70b
RAM: 8GB minimum for 8B model, 64GB for 70B model
GPU: Optional but highly recommended (10x+ faster with GPU)

Recommended for your use case:

Start with llama3.1:8b - fast, runs on most machines
If accuracy isn't perfect, try llama3.1:70b (needs more resources)
Or try llama3.2:3b if you have limited RAM

forsyth2:
@chengzhuzhang I wasn't able to complete the AI-component during the datathon, but I was able to create fairly- accurate starter configs using two deterministic scripts. These may be useful to give as context to your AI config builder. I think it's possible to give AI agents specific code blocks to run as well.

> 1. A script to extract available data from a simulation output directory. That's `simulation_output_reviewer.py`.
> 2. A script to generate a starting point cfg file based on the data found. That's `zppy_config_generator.py`.

chengzhuzhang:
@forsyth2 thanks for pointing this out. I like the idea of giving agents confined code blocks to run as a way to better guide them. I’ll take a closer look.

PR #806, opened 2026-04-10
Fixes of unified rc4 and rc5

forsyth2:
## Summary

Objectives:
- Addressing issues in Unified 1.13.0rc4
  - Commit 1: Fix error codes to use logical number sequence (and include one after the `zi-global-time-series` call)
  - Commits 2-4: Standardizing mapping file paths
- Addressing issues in Unified 1.13.0rc5
  - Commits 5 & 10: Log the package version
  - Commit 6: Add more status logging to `tc_analysis`
  - Commit 7: Increase diags walltime on Compy
  - Commit 8: Fix to variables feeding into `e3sm_to_cmip`
  - Commit 9: Add ability to test arbitrary/developer version of `e3sm_to_cmip`

Issue resolution:
- Resolves issues encountered during:
  - [Compy test of E3SM Unified 1.13.0rc4](https://github.com/E3SM-Project/zppy/discussions/802#discussioncomment-16508424)
  - [Compy test of E3SM Unified 1.13.0rc5](https://github.com/E3SM-Project/zppy/discussions/802#discussioncomment-16560117)
  - [Compy follow-up limited test](https://github.com/E3SM-Project/zppy/discussions/802#discussioncomment-16577319)
  - [Perlmutter follow-up limited test](https://github.com/E3SM-Project/zppy/discussions/802#discussioncomment-16607020)

Select one: This pull request is...
- [x] a bug fix: increment the patch version
- [ ] a small improvement: increment the minor version
- [ ] a new feature: increment the minor version
- [ ] an incompatible (non-backwards compatible) API change: increment the major version

## Small Change

- [ ] To merge, I will use "Squash and merge". That is, this change should be a single commit.
   - However, I will use a merge commit since each commit does clearly distinct work.
- [x] Logic: I have visually inspected the entire pull request myself.
- [x] Pre-commit checks: All the pre-commits checks have passed.

PR #810, opened 2026-04-16
Replace `ncclimo_cmd` with `nco_path`

### LINKED ISSUE #809: [Feature]: Support different NCO versions for calls other than `ncclimo`

#761 added the `ncclimo_cmd` parameter so we could easily switch out NCO versions, usually meaning switching between `ncclimo` (regular/production) and `/home/ac.zender/bin_chrysalis/ncclimo --npo` (testing latest NCO code).

However, `ncclimo` is far from the only NCO command. As part of [E3SM Unified 1.13.0rc5 testing](https://github.com/E3SM-Project/zppy/discussions/802#discussioncomment-16575468), it is instead `ncremap` that needs an updated NCO version (using `/global/cfs/cdirs/e3sm/zender/bin/ncremap --npo`)

We have three possible options:

1. Ideally, NCO would support development environments, similar to how `e3sm_to_cmip`, `e3sm_diags`, `mpas_analysis`, and `zppy-interfaces` specify a dev environment setup in files like `conda/dev.yml`. This is the simplest approach. Both `climo.bash` and `ts.bash` already include `{{ environment_commands }}`, so there actually wouldn't be any user-facing code changes on the `zppy` side. For `zppy`'s testing, we could easily replicate for NCO what we do for the packages listed above. For instance, [this commit](https://github.com/E3SM-Project/zppy/pull/806/changes/fd6398cfb05667a26c495e83d7bc37f731cfe31f) enables the tests to set a distinct `environment_commands` for the `e3sm_to_cmip` package.
2. Every time a new NCO command needs a specific version installed, add a new parameter to `defaults/default.ini` (For instance, `ncremap_cmd` would join `ncclimo_cmd`). I dislike this option because it could quickly increase the number of parameters in the `zppy` `cfg`. It means essentially repeating #761 for every NCO command that `zppy` uses, not to mention needing to replace every single NCO command. I.e., in #761 every call of `ncclimo` became `{{ ncclimo_cmd }}` instead.
3. Deprecate the `ncclimo_cmd` option and replace it with something like `use_dev_nco=string(default="false")`, and then at the top of every `.bash` template using a NCO command, include an if/else block that sets the proper command for every NCO command used in that file (or to avoid code duplication, I think this could be set just once by doing it on the Python side). The advantage over (2) is that fewer parameters are introduced, but in both (2) & (3), we can't get around replacing every single NCO call with something like `{{ ncclimo_cmd }}`. Only option (1) allows us to just continue calling the NCO commands as they are.

_No response_

### Additional context

_No response_

forsyth2:
## Summary

Objectives:
-  Replace limited-use `ncclimo_cmd` with more polymorphic parameter `nco_path` that can be used for any and all NCO commands.

Issue resolution:
- Closes #809

## Small Change

   - I will use a merge commit though, since the commits here each do clearly distinct chunks of work.


forsyth2:
This pull request is based on top of #806's branch (which currently accounts for the first 9 commits). I decided to implement the `nco_path` parameter as a separate PR since it was a big enough piece of work.

For code reviewers, the [next 3 commits](https://github.com/E3SM-Project/zppy/pull/810/changes/fd6398cfb05667a26c495e83d7bc37f731cfe31f..72fc7746c4e2f91c56387307a215f83ba3f326b1) are what's relevant.

The [final commit](https://github.com/E3SM-Project/zppy/pull/810/changes/699701c437bb41573bc53b2ec701117991460965) is entirely testing-side, analogous to the [e3sm_to_cmip_environment_commands addition](https://github.com/E3SM-Project/zppy/pull/806/changes/fd6398cfb05667a26c495e83d7bc37f731cfe31f).

zppy/defaults/default.ini
```diff
@@ -46,11 +46,15 @@ input_subdir = string(default="archive/atm/hist")
...
-# Which will use the latest (i.e., unreleased) version of ncclimo (on Chrysalis).
+# DEPRECATED. No longer used.
 ncclimo_cmd = string(default="ncclimo")
+# Keep as the empty string to use production-version NCO commands
+# Set to the development-version path to use that instead.
+# Example (LCRC): Set `/home/ac.zender/bin_chrysalis`
+# Then, `ncclimo` would become `/home/ac.zender/bin_chrysalis/ncclimo --npo`
+# Example (NERSC): Set `/global/cfs/cdirs/e3sm/zender/bin`
+# Then, `ncclimo` would become `/global/cfs/cdirs/e3sm/zender/bin/ncclimo --npo`
+nco_path = string(default="")
```
forsyth2, line 57: We've replaced `ncclimo_cmd` with `nco_path`

zppy/templates/inclusions/boilerplate.bash
```diff
@@ -13,3 +13,24 @@ id=${SLURM_JOBID}
...
+  local environment_path="{{ nco_path }}"
+  local nco_cmd=$1
+  shift # Remove nco_cmd from the argument list
+
+  if [[ -z "$environment_path" ]]; then
+    # nco_path is empty, so use just the nco_cmd itself:
+    "$nco_cmd" "$@"
+  else
+    # nco_path is non-empty, so use the full development setup:
+    "${environment_path%/}/$nco_cmd" --npo "$@"
```
forsyth2, line 47: This should cover basically all development-version cases, correct?

forsyth2 (commented):
@czender @chengzhuzhang Could you please do at least a visual inspection code review of [these 3 commits](https://github.com/E3SM-Project/zppy/pull/810/changes/fd6398cfb05667a26c495e83d7bc37f731cfe31f..72fc7746c4e2f91c56387307a215f83ba3f326b1)? Thanks!

zppy/defaults/default.ini
```diff
@@ -46,11 +46,15 @@ input_subdir = string(default="archive/atm/hist")
...
-# Default is simply "ncclimo"
-# The most likely alternative option is `/home/ac.zender/bin_chrysalis/ncclimo --npo`,
-# Which will use the latest (i.e., unreleased) version of ncclimo (on Chrysalis).
+# DEPRECATED. No longer used.
 ncclimo_cmd = string(default="ncclimo")
+# Keep as the empty string to use production-version NCO commands
+# Set to the development-version path to use that instead.
+# Example (LCRC): Set `/home/ac.zender/bin_chrysalis`
+# Then, `ncclimo` would become `/home/ac.zender/bin_chrysalis/ncclimo --npo`
+# Example (NERSC): Set `/global/cfs/cdirs/e3sm/zender/bin`
```
chengzhuzhang: I think the NERSC path currently is /global/cfs/cdirs/e3sm/zender/bin_perlmutter

forsyth2: Hmm they look identical. @czender Does it matter? You noted `/global/cfs/cdirs/e3sm/zender/bin/` [here](https://e3sm-project.slack.com/archives/C06R7PW4DL0/p1776368033589389?thread_ts=1776364484.730299&cid=C06R7PW4DL0)

```
> ls /global/cfs/cdirs/e3sm/zender/bin_perlmutter/
antlr         hello   ncap2    ncbo       ncclimo  ncea    nces     ncks   ncra    ncremap   ncwa     sng
antlr-config  makdep  ncatted  ncchecker  ncdiff   ncecat  ncflint  ncpdq  ncrcat  ncrename  ncz2psx

> ls /global/cfs/cdirs/e3sm/zender/bin
antlr         hello   ncap2    ncbo       ncclimo  ncea    nces     ncks   ncra    ncremap   ncwa     sng
antlr-config  makdep  ncatted  ncchecker  ncdiff   ncecat  ncflint  ncpdq  ncrcat  ncrename  ncz2psx
```

forsyth2: Oh it looks like `/global/cfs/cdirs/e3sm/zender/bin` is actually a link to `/global/cfs/cdirs/e3sm/zender/bin_perlmutter/`

zppy/templates/inclusions/boilerplate.bash
chengzhuzhang: this looks okay to me.

zppy/defaults/default.ini
czender: The `bin_perlmutter` path is safer since if/when NERSC cross-mounts a new HPC it will have its own execution directory `bin_newmachine` that might need to co-exist with PM executables.

zppy/templates/inclusions/boilerplate.bash
```diff
@@ -13,3 +13,24 @@ id=${SLURM_JOBID}
 # Update status file
 STARTTIME=$(date +%s)
 echo "RUNNING ${id}" > {{ prefix }}.status
+
+run_nco() {
```
czender, line 32: I can't be sure but it looks to me like `run_nco` might be invoked on operator besides `ncclimo` and `ncremap`, yet those are the only two operators that suppor the `--npo` syntax. None of the binary executable operators (ncra, ncwa, ncap2...) support `--npo`. Testing those operators would be done by prefixing them with the absolute path to the development executable directory.

forsyth2: > might be invoked on operator besides

Yes, that's correct. [This commit](https://github.com/E3SM-Project/zppy/pull/810/changes/72fc7746c4e2f91c56387307a215f83ba3f326b1) introduces that. Ok, so that needs to be changed? 

> Testing those operators would be done by prefixing them with the absolute path to the development executable directory.

@czender I'm confused. Are you saying for `ncclimo` and `ncremap`, we should have `"${environment_path%/}/$nco_cmd" --npo "$@"` but for everything else we should have only `"${environment_path%/}/$nco_cmd" "$@"`?

czender:
I'm a bit confused so it's no wonder you're confused too. The new architecture looks good for `ncremap` and `ncclimo` only. There is currently no one-line-change way to test development versions of the binary executables. So forget what I said above about testing them with the development executable path prepended. When testing binaries the LD_LIBRARY_PATH would also need to be modified (in addition to the executable path) so that the binaries would see the development libary in my directory rather than the older E3SMU NCO library. That would be a much larger change (it's essentially what `--npo` does for the scripts). In summary there's no way to test the stand-alone binaries unless you also want to implement a way to ensure the binaries see the right library.

forsyth2:
> There is currently no one-line-change way to test development versions of the binary executables.
> it's essentially what `--npo` does for the scripts

So `ncclimo` and `ncremap` are the only scripts? Everything else is a binary executable?

I'll the update the logic of `run_nco` accordingly.

forsyth2:
I added a [commit](https://github.com/E3SM-Project/zppy/pull/810/changes/fabbb57826452f9ab070e959d56ab529f47a532d) addressing these comments

zppy/templates/inclusions/boilerplate.bash
```diff
@@ -16,21 +16,23 @@ echo "RUNNING ${id}" > {{ prefix }}.status
...
-    # nco_path is empty, so use just the nco_cmd itself:
-    "$nco_cmd" "$@"
-  else
-    # nco_path is non-empty, so use the full development setup:
+  if [[ -n "$environment_path" && ( "$nco_cmd" == "ncclimo" || "$nco_cmd" == "ncremap" ) ]]; then
+    # nco_path is non-empty AND nco_cmd is a script (either ncclimo or ncremap),
+    # so use the full development setup:
     "${environment_path%/}/$nco_cmd" --npo "$@"
+  else
+    # nco_path is empty OR nco_cmd is a binary executable (not a script),
```
czender, line 49: Yes, I think this logic will work.

czender (commented):
Yes ncclimo and ncremap are the only scripts. I think the new fix should work.

forsyth2:
Rebased off `main`, after merging #806

forsyth2:
This was tested [here](https://github.com/E3SM-Project/zppy/discussions/802#discussioncomment-16607020), so I think this should be good to merge now.

